PR #24, opened 2020-08-11
Expose Tap ring size in env

sknat:
both calico-agent and vpp-manager now take `CALICOVPP_TAP_RING_SIZE` that can be either 
* an int specifying the tap ring buffer size (both RX&TX)
* two ints (RxSize,TxSize) separated by a comma
Default size is 1024


calico-vpp-agent/cni/network_vpp.go
```diff
@@ -414,6 +414,8 @@ func (s *Server) AddVppInterface(args *pb.AddRequest, doHostSideConf bool) (ifNa
 		MacAddress:     vppSideMacAddress,
 		HostMacAddress: containerSideMacAddress,
 		RxQueues:       config.TapRXQueues,
+		RxRingSize:     config.TaxRxRingSize,
+		TxRingSize:     config.TaxTxRingSize,
```
AloysAugustin, line 418: Typo?

sknat: Indeed, didn't see it at first glance, thanks 😄

sknat: I'll do the taxreform then 😛

AloysAugustin: 😃

PR #37, opened 2020-08-20
Fix build env

AloysAugustin:


AloysAugustin:
This saves time on the build, and avoids a potential hang if there are some updates to be installed by install-dep. The only drawback is the need to update the Dockerfile when the VPP dependencies change, which is very infrequent - and only happens when we change the VPP commit we use, meaning we need to rebuild an image to test it, and have to notice that a change is needed.

sknat (approved):

PR #39, opened 2020-08-25
Remove need for fixed node names in the tests

AloysAugustin:


test/scripts/test.sh
```diff
@@ -57,15 +57,26 @@ calico_build_nptest ()
   make docker DOCKERREPO=calicovpp/nptest
 }
 
+get_nodes ()
+{
+  NODES=($(kubectl get nodes -o jsonpath="{.items[*].metadata.name}"))
+  if [ ${#NODES[@]} -lt 2 ]; then
```
sknat, line 63: If possible issuing a warning & setting `NODE_2 = NODE_1` would be good.
I have some tests (e.g. natOutgoing running with a single node)

AloysAugustin: Ok, this should be fixed in the last commit.

sknat (approved):

PR #67, opened 2020-11-05
Feature/improvements

sknat:
- Improve recovery of ipip tunnels
- RR pinning of tun rx queues to VPP workers

### DIFF
--- a/calico-vpp-agent/cni/network_vpp.go
+++ b/calico-vpp-agent/cni/network_vpp.go
@@ -306,10 +306,10 @@ func (s *Server) delVppInterfaceHandleRoutes(swIfIndex uint32, isIPv6 bool) erro
 			}
 		}
 
-		s.log.Warnf("vpp del route %s", route.String())
+		s.log.Infof("Delete VPP route %s", route.String())
 		err = s.vpp.RouteDel(&route)
 		if err != nil {
-			s.log.Warnf("vpp del route %s err: %v", route.String(), err)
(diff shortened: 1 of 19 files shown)


calico-vpp-agent/cni/network_vpp.go
```diff
@@ -214,6 +214,16 @@ func (s *Server) AddVppInterface(podSpec *LocalPodSpec, doHostSideConf bool) (sw
 	}
 	s.log.Infof("created tun[%d]", swIfIndex)
 
+	if tun.NumRxQueues > 1 {
+		for i := 0; i < tun.NumRxQueues; i++ {
+			worker := uint32(tun.NumRxQueues) % s.NumVPPWorkers
```
AloysAugustin, line 219: This seems to put all the rx queues on the same worker, is this really what we need?

sknat: Ah yes, good catch, it's `i % Nworkers`

AloysAugustin: Actually `((tun.SwIfIndex * tun.NumRxQueues) + i) % s.NumVppWorkers` could be even better, otherwise you won't use all workers if you have more workers than queues configured. This configuration may be interesting in high-density clusters...

sknat: Right !

AloysAugustin (approved):

PR #69, opened 2020-11-06
image: respect CALICO_DISABLE_FILE_LOGGING for calico agent

AloysAugustin:
@sknat  Do you see any other places that should be updated to take into account the new default log output?


sknat:
Yes, updated it in vppdev.sh

sknat (approved):

PR #88, opened 2020-11-26
fix/node snat

AloysAugustin:


calico-vpp-agent/routing/connectivity/ipip.go
```diff
@@ -46,10 +46,8 @@ func (p *IpipProvider) RescanState() {
 	nodeIP4 := p.getNodeIP(false)
 	nodeIP6 := p.getNodeIP(true)
 	for _, tunnel := range tunnels {
-		if tunnel.Src.Equal(nodeIP4) || tunnel.Src.Equal(nodeIP6) {
```
sknat, line 49: Not sure if we should remove this. Is it related to some bug ?

sknat: -> broken because we don't get the nodeIP[46] soon enough. Will fix later on

sknat (approved):

PR #92, opened 2020-12-04
add config for Tap MTU

MathiasRaoul:
Signed-off-by: MathiasRaoul <mathias.raoul@gmail.com>


yaml/overlays/dev/kustomize.sh
```diff
@@ -194,6 +194,7 @@ calico_create_template ()
   export CALICOVPP_VPP_STARTUP_SLEEP=${CALICOVPP_VPP_STARTUP_SLEEP:=0}
   export CALICOVPP_TAP_RX_QUEUES=${CALICOVPP_TAP_RX_QUEUES:=1}
   export CALICOVPP_TAP_TX_QUEUES=${CALICOVPP_TAP_TX_QUEUES:=1}
+  export CALICOVPP_TAP_MTU=${CALICOVPP_TAP_MTU:=0}
```
AloysAugustin, line 197: Same here we should default to 1440

calico-vpp-agent/config/config.go
```diff
@@ -75,6 +76,7 @@ var (
 	ServiceCIDRs      []*net.IPNet
 	TapRxQueueSize    int = 0
 	TapTxQueueSize    int = 0
+	TapMtu            int = 0
```
AloysAugustin, line 79: If this is the default for the agent, we should configure this to 1440

yaml/overlays/dev/calicodev.yaml
```diff
@@ -284,6 +290,12 @@ spec:
                 configMapKeyRef:
                   name: calico-config
                   key: CALICOVPP_TAP_RING_SIZE
+            - name: CALICOVPP_TAP_MTU
```
AloysAugustin, line 293: You can remove this since we will use the host's original MTU for the vpp-manager tap

vpp-manager/startup/startup.go
```diff
@@ -201,6 +203,15 @@ func parseEnvVariables(params *config.VppManagerParams) (err error) {
 		}
 	}
 
+	params.TapMtu = DefaultTapMtu
```
AloysAugustin, line 206: For vpp-manager we should use the MTU that we find on the uplink interface originally instead of configuring it through an env var. You should modify getInterfaceConfigFromLinux to grab the MTU from link, save it in the InterfaceConfig struct, and then use that to configure the tap.

AloysAugustin (changes requested):

yaml/overlays/dev/kustomize.sh
sknat: This might be debatable, `0` will call VPP's api with `mtu_set = false` so afaik use linux's default (which in the end might be 1500). Depends on what's best for most usecases.

yaml/overlays/dev/calicodev.yaml
AloysAugustin: As discussed offline, let's actually keep this as an option to override the default computed MTU for vpptap0.

sknat (approved):

AloysAugustin (approved):

PR #112, opened 2021-01-22
Fix AVF uplink & broken pipe

sknat:
* When using avf uplink, vpp-manager will now :
  * if the interface is a PF without a VF, create a VF + generate a mac for it. Addresses will be taken on the PF in linux
  * if the interface is a PF with an existing VF, it will use the first vf and its mac. ip addresses & routes taken on the linux PF
  * if the interface is a VF, it will use it as is
* This also fixes the broken pipe signal being mishandled & forwarded to VPP

### DIFF
--- a/vpp-manager/main.go
+++ b/vpp-manager/main.go
@@ -90,7 +90,7 @@ func handleSignals() {
 			} else {
 				log.Infof("ignoring sigchld")
 			}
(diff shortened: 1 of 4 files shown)


AloysAugustin (approved):
Good approach :)

PR #114, opened 2021-01-27
agent : soustract max encap header size from configured tap mtu

MathiasRaoul:


calico-vpp-agent/cni/cni_server.go
```diff
@@ -232,6 +251,9 @@ func NewServer(v *vpplink.VppLink, rs *routing.Server, ps *policy.Server, l *log
 		l.Fatalf("failed to listen on %s: %v", config.CNIServerSocket, err)
 		return nil, err
 	}
+
+	vppLinuxMtu, err := fetchVppLinuxMtu()
+
```
sknat, line 256: Might be worth logging the error

calico-vpp-agent/cni/network_vpp.go
```diff
@@ -203,6 +203,16 @@ func (s *Server) AddVppInterface(podSpec *storage.LocalPodSpec, doHostSideConf b
 		return swIfIndex, nil
 	}
 
+	// configure MTU from env var if present or calculate it from host mtu
+	var tapMtu int = 0
+	if config.TapMtu > 0 {
```
sknat, line 208: We also need to change the default config.TapMtu value (when `CALICOVPP_TAP_MTU` is not set), in `config.go:80`

sknat (changes requested):

sknat (approved):

PR #120, opened 2021-02-02
yaml: disable ping plugin by default

sknat:


AloysAugustin (changes requested):
We should change the dpdk and eks-dpdk overlays as well

sknat:
Right, went a bit quick on that one

AloysAugustin (approved):

PR #121, opened 2021-02-03
vpp-manager: enable GSO on vpptap0

sknat:
Add a `CALICOVPP_ENABLE_GSO` switch to vpp-manager that enables GSO in vpp on the PHY & the vpptap0 tun interface punting packets to the host.


AloysAugustin:
Do we really want to add the possibility to disable GSO on the phy? This creates the possibility for users to write an invalid config if they enable GSO on the pods but not on the phy. It may be better to leave GSO always enabled on the phy, and use this config parameter only for vpptap0 ?

sknat:
I agree this isn't ideal as we have no way to use the same config parameter for 'enable pod GSO' & 'enable tap GSO', but I think it's worth having a way to globally disable GSO on the cluster, if only for the purpose of debugability.
Then there is also the question of weather we allow for setting GSO on/off separately on the PHY & vpptap0

AloysAugustin (approved):
LGTM with the name changes

PR #123, opened 2021-02-03
vpp-manager: rm phy routes&addr before vpp

sknat:

### DIFF
--- a/vpp-manager/uplink/common.go
+++ b/vpp-manager/uplink/common.go
@@ -72,13 +72,15 @@ func (d *UplinkDriverData) removeLinuxIfConf(setIfDown bool) {
 		for _, addr := range d.conf.Addresses {
 			err := netlink.AddrDel(link, &addr)
 			if err != nil {
-				log.Errorf("Error adding address %s to tap interface : %+v", addr, err)
+				log.Errorf("Error removing address %s from tap interface : %+v", addr, err)
 			}
 		}
 
 		if d.conf.IsUp && setIfDown {
 			err = netlink.LinkSetDown(link)
 			if err != nil {
+				// In case it still succeeded
+				netlink.LinkSetUp(link)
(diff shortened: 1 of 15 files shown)


vpp-manager/uplink/common.go
```diff
@@ -55,6 +55,36 @@ func (d *UplinkDriverData) GetName() string {
...
+		for _, addr := range d.conf.Addresses {
+			err := netlink.AddrDel(link, &addr)
+			if err != nil {
+				log.Errorf("Error adding address %s to tap interface : %+v", addr, err)
+			}
+		}
+
+		if d.conf.IsUp && setIfDown {
+			err = netlink.LinkSetDown(link)
+			if err != nil {
```
AloysAugustin, line 81: In SafeSetInterfaceDownByName , if LinkSetDown errored we attempted to put the link back up, should we do it here as well?

vpp-manager/uplink/common.go
```diff
@@ -55,6 +55,36 @@ func (d *UplinkDriverData) GetName() string {
...
+			err := netlink.RouteDel(&route)
+			if err != nil {
+				log.Errorf("cannot delete route %+v: %+v", route, err)
+			}
+		}
+		// Remove addresses to not have them conflict with vpptap0
+		for _, addr := range d.conf.Addresses {
+			err := netlink.AddrDel(link, &addr)
+			if err != nil {
+				log.Errorf("Error adding address %s to tap interface : %+v", addr, err)
```
AloysAugustin, line 75: The log should be changed (adding -> removing)

vpp-manager/uplink/common.go
sknat: Right, yesterday I didn't recall why we did have it. But on a second thought it makes sense

vpp-manager/uplink/common.go
sknat: right

AloysAugustin (approved):

PR #137, opened 2021-03-17
Ipsec Async crypto

MathiasRaoul:


calico-vpp-agent/routing/connectivity/ipsec.go
```diff
@@ -34,6 +34,29 @@ type IpsecProvider struct {
 }
 
 func (p *IpsecProvider) OnVppRestart() {
+	var err error
+	err = p.vpp.SetIPsecAsyncMode(true)
```
sknat, line 38: Maybe we should do this only if `config.IpsecNbAsyncCryptoThread > 0` so that we can also test with async crypto off ?

calico-vpp-agent/cni/network_vpp.go
```diff
@@ -234,12 +234,25 @@ func (s *Server) AddVppInterface(podSpec *storage.LocalPodSpec, doHostSideConf b
 	}
 	s.log.Infof("created tun[%d]", swIfIndex)
 
-	if tun.NumRxQueues > 1 && s.NumVPPWorkers > 0 {
+	var nbDataThread int = 0
+	if config.IpsecNbAsyncCryptoThread > 0 {
```
sknat, line 238: Maybe we should put the `if nbDataThread <= 0` condition only inside the async crypto `if`. afaik `nbDataThread==0` is legal when `workers==0`
````go
nbDataThread := (int) s.NumVPPWorkers
if config.IpsecNbAsyncCryptoThread > 0 {
       nbDataThread = nbDataThread - config.IpsecNbAsyncCryptoThread
       s.log.Infof("ipsec async crypto %d", config.IpsecNbAsyncCryptoThread)
       if nbDataThread <= 0 {
     		s.log.Error("not enough thread for async crypto")
   		nbDataThread = (int) s.NumVPPWorkers
	}
}
````

sknat (approved):

Issue #138, opened 2021-03-19
Host DNS broken after installing calico-vpp


lwr20:
**Environment**
- Calico/VPP version: 0.13.0
- Kubernetes version: 1.20
- Deployment type: AWS VMs, tried Ubuntu 18.04 and Ubuntu 20.04, installed using kubeadm.
- Network configuration: Used https://raw.githubusercontent.com/projectcalico/vpp-dataplane/release/\
v0.13.0-calicov3.17.1/yaml/generated/calico-vpp-nohuge.yaml .  Set VPP interface to ens5 and pod-cidr to 192.168.0.0/16.

**Issue description**
After applying calico-vpp-nohuge.yaml, the nodes were unable to do DNS lookups, so core-dns and calico-kube-controllers pods were stuck in ImagePullBackoff.

**To Reproduce**
- Setup cluster
- kubectl apply -f calico-vpp-nohuge.yaml

**Expected behavior**
I'd expect core-dns and calico-kube-controllers pods to be pulled correctly.

**Additional context**
Discussed with Aloys in slack.  

Running `sudo tcpdump -v -n -i vpptap0 port 53` I don't see anything when I run `nslookup www.google.com` on one of my hosts.  The nslookup gets
```
** server can't find www.google.com: SERVFAIL
```
instantly (no delay or timeout)

This works though:
```
ubuntu@ip-172-16-101-168:~$ nslookup microsoft.com 172.16.0.2
Server:		172.16.0.2
Address:	172.16.0.2#53
Non-authoritative answer:
Name:	microsoft.com
Address: 40.112.72.205
Name:	microsoft.com
Address: 40.113.200.201
Name:	microsoft.com
Address: 13.77.161.179
Name:	microsoft.com
Address: 104.215.148.63
Name:	microsoft.com
Address: 40.76.4.15
```
> Aloys Augustin  2 days ago
Could you try running :
`sudo systemd-resolve --interface vpptap0 --set-dns 172.16.0.2`
If that solves the issue, I’ll look into how we can do it automatically when configuring the interface

And that did indeed solve the issue.

AloysAugustin:
[...]

Issue #139, opened 2021-03-19
issue in the af_packet driver in vpp on VM with 9K MTU


lwr20:
**Environment**
- Calico/VPP version: 0.13.0
- Kubernetes version:  1.20
- Deployment type: AWS VMs, Ubuntu 20.04.  Installed using kubeadm.
- Network configuration: Applied https://raw.githubusercontent.com/projectcalico/vpp-dataplane/release/v0.13.0-calicov3.17.1/yaml/generated/calico-vpp-nohuge.yaml - vpp interface was set to ens5 and pid-cidr to 192.168.0.0/16.
 
**Issue description**
On VMs with 9K MTU, calico-kube-controllers stuck in ImagePullBackoff
kubectl describe on the pod showed:
```
│   Normal   Scheduled  90s                default-scheduler                                      Successfully assigned kube-system/calico-kube-controllers-744cfdf676-lkjdc to ip │
│ -172-16-101-204.us-west-2.compute.internal                                                                                                                                       │
│   Warning  Failed     88s                kubelet, ip-172-16-101-204.us-west-2.compute.internal  Failed to pull image "docker.io/calico/kube-controllers:v3.17.1": rpc error: cod │
│ e = Unknown desc = error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/27/278f40d9f3b82fdb687ddc52dda216 │
│ 82525bd4c814fa3ab670fa44df150a1252/data?verify=1616002988-RdO%2Bh7V4yFIOUJHRch%2F%2FNMyoGKs%3D: tls: first record does not look like a TLS handshake                             │
│   Warning  Failed     74s                kubelet, ip-172-16-101-204.us-west-2.compute.internal  Failed to pull image "docker.io/calico/kube-controllers:v3.17.1": rpc error: cod │
│ e = Unknown desc = error pulling image configuration: Get https://production.cloudflare.docker.com/registry-v2/docker/registry/v2/blobs/sha256/27/278f40d9f3b82fdb687ddc52dda216 │
│ 82525bd4c814fa3ab670fa44df150a1252/data?verify=1616003002-yZ27PHtgHqCDMSsBdYrcJtgUvw4%3D: tls: first record does not look like a TLS handshake                                   │
[...]

PR #142, opened 2021-03-30
Use original interface name and mac for host connectivity

AloysAugustin:


AloysAugustin:
This should fix #138 and #139

yaml/base/calico-vpp.yaml
```diff
@@ -74,6 +74,7 @@ data:
       full-coredump
       cli-listen /var/run/vpp/cli.sock
       pidfile /run/vpp/vpp.pid
+      exec /etc/vpp/startup.exec
```
sknat, line 77: Might be added for tests no ?

yaml/overlays/test-vagrant-mounts/mounts.yaml
```diff
@@ -24,6 +24,10 @@ spec:
               mountPath: /bin/calico-vpp-agent
               readOnly: true
         - name: vpp
+          # command:
```
sknat, line 27: Just rogue comments :)

yaml/base/calico-vpp.yaml
AloysAugustin: I added it there because it makes one less thing to configure when we need it (for instance for some debugging), and it doesn't hurt to have it if it's empty.

yaml/overlays/test-vagrant-mounts/mounts.yaml
AloysAugustin: Good catch, removed :)

yaml/base/calico-vpp.yaml
sknat: Right, works :)

sknat (approved):

PR #184, opened 2021-05-27
RDMA driver

AloysAugustin:


vpp-manager/uplink/common.go
```diff
@@ -167,11 +168,16 @@ func SupportedUplinkDrivers(params *config.VppManagerParams, conf *config.Interf
 	if d := NewAFPacketDriver(params, conf); d.IsSupported(false /* warn */) {
 		lst = append(lst, d)
 	}
+	if d := NewRDMADriver(params, conf); d.IsSupported(false /* warn */) {
```
AloysAugustin, line 171: The order is important in this function, drivers that we put first in the list will be tried first. Since RDMA is a native driver, it will be the most efficient option to drive an interface it supports, so we want to put it near the top of the list (next to virtio and avf).

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type RDMADriver struct {
+	UplinkDriverData
+}
+
+func (d *RDMADriver) IsSupported(warn bool) (supported bool) {
+	var ret bool
+	supported = true
+	ret = d.params.LoadedDrivers[config.DRIVER_VFIO_PCI]
```
AloysAugustin, line 35: I think vfio / uio drivers are not required for this driver

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+	if !ret && warn {
+		log.Warnf("did not find vfio-pci or uio_pci_generic driver")
+		log.Warnf("VPP may fail to grab its interface")
+	}
+	supported = supported && ret
+
+	ret = d.params.AvailableHugePages > 0
```
AloysAugustin, line 42: According to Ben the RDMA driver doesn't require hugepages, so we can remove this check as well.

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+	if !ret && warn {
+		log.Warnf("No hugepages configured, driver won't work")
+	}
+	supported = supported && ret
+
+	ret = d.conf.Driver == config.DRIVER_VIRTIO_PCI
```
AloysAugustin, line 48: This probably need to be updated to the actual driver for mlx interfaces.

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	supported = supported && ret
+
+	return supported
+}
+
+func (d *RDMADriver) PreconfigureLinux() (err error) {
+	newDriverName := d.params.NewDriverName
+	doSwapDriver := d.conf.DoSwapDriver
+	/*if newDriverName == "" {
```
AloysAugustin, line 60: Please remove everything related to driver swap since it is not required for this driver.

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (d *RDMADriver) RestoreLinux() {
+	if !d.params.VfioUnsafeiommu {
+		err := utils.SetVfioUnsafeiommu(false)
+		if err != nil {
+			log.Warnf("Virtio restore error %v", err)
+		}
+	}
+	if d.conf.PciId != "" && d.conf.Driver != "" {
```
AloysAugustin, line 88: Same as before, this is not required for this driver.

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+		err = utils.SwapDriver(d.conf.PciId, newDriverName, true)
+		if err != nil {
+			log.Warnf("Failed to swap driver to %s: %v", newDriverName, err)
+		}
+	}
+	return nil
+}
+
+func (d *RDMADriver) RestoreLinux() {
+	if !d.params.VfioUnsafeiommu {
```
AloysAugustin, line 82: Since we don't use the vfio kernel driver, this is not required.

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,139 @@
...
+	d.params.NumRxQueues,
+	d.params.RxQueueSize,
+	d.params.TxQueueSize,
+	}
+	swIfIndex, err := vpp.CreateRDMA(MyInfo)
+
+	if err != nil {
+		return errors.Wrapf(err, "Error creating RDMA interface")
+	}
+	err = d.moveInterfaceToNS(d.params.MainInterface, vppPid)
```
AloysAugustin, line 121: Please check the value of err here and return an error if moving the interface fails.

yaml/overlays/test-vagrant-mounts-flat/flat.yaml
```diff
@@ -56,10 +56,10 @@ spec:
       containers:
         - name: vpp
           env:
-            - name: CALICOVPP_CONFIG_EXEC_TEMPLATE
-              value: create interface rdma host-if __VPP_DATAPLANE_IF__ rx-queue-size 1024 tx-queue-size 1024 num-rx-queues 4
-            - name: CALICOVPP_VPP_STARTUP_SLEEP
-              value: "5"
+            #- name: CALICOVPP_CONFIG_EXEC_TEMPLATE
```
AloysAugustin, line 59: Please remove comments entirely.

AloysAugustin (commented):
Thanks @hedibouattour , this looks great, I left a few minor comments in the code.
Also, please ensure your editor is configured to run gofmt on save, I think there are a few places where the code is not formatted correctly. In VSCode I think it should be sufficient to install the official go extensions.

vpplink/rdma.go
```diff
@@ -0,0 +1,50 @@
...
+	TxqSize int
+}
+
+func (v *VppLink) CreateRDMA(rdmainfo RDMAInfo) (swIfIndex uint32, err error) {
+	v.lock.Lock()
+	defer v.lock.Unlock()
+	response := &rdma.RdmaCreateV2Reply{}
+	request := &rdma.RdmaCreateV2{
+		HostIf:  rdmainfo.HostIf,
+		Name:    "RDMA-",
```
sknat, line 38: Maybe we could use VPP's default (so not passing `Name` at all)

vpplink/rdma.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type RDMAInfo struct {
+	HostIf  string
+	RxqNum  int
+	RxqSize int
+	TxqSize int
+}
+
+func (v *VppLink) CreateRDMA(rdmainfo RDMAInfo) (swIfIndex uint32, err error) {
```
sknat, line 32: Might be worth making `rdmainfo` a pointer, also maybe worth using a name like `RDMAInterface`
At some point, we could even try to make the all the Interface structs uniforms (and share a common struct)

vpplink/rdma.go
```diff
@@ -0,0 +1,50 @@
...
+package vpplink
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/projectcalico/vpp-dataplane/vpplink/binapi/vppapi/rdma"
+)
+
+type RDMAInfo struct {
```
sknat, line 25: Usually putting these in the `types` package, not sure if it really makes sense after all, 
but we can keep the logic for now

vpp-manager/uplink/rdma.go
```diff
@@ -0,0 +1,100 @@
...
+		log.Warnf("Error setting %s up: %v", d.params.MainInterface, err)
+		return
+	}
+
+	// Re-add all adresses and routes
+	d.restoreLinuxIfConf(link)
+}
+
+func (d *RDMADriver) CreateMainVppInterface(vpp *vpplink.VppLink, vppPid int) (err error) {
+	MyInfo := vpplink.RDMAInfo{
```
sknat, line 68: Same as below, maybe use a name like `RdmaInterface`

sknat (changes requested):
Nice 😄  
Just a few minor quirks

sknat (approved):

Issue #141, opened 2021-03-25
Adding support to SRv6 


zvfvrv:
**Is your feature request related to a problem? Please describe.**
This is not a problem but a new feature: use of a new overlay type in addition to IPinIP and VXLAN, called SRv6 which is based on Segment Routing over IPv6 

**Describe the solution you'd like**
We want to add an additional overlay type called SRv6 in addition to IPinIP and VXLAN.

**Describe alternatives you've considered**
This is a new feature, which is an alternative to IPinIP and VXLAN.

**Additional context**
SRv6 is a networking solution based on RFC [8754](https://tools.ietf.org/html/rfc8754) and on the recent RFC [8986](https://tools.ietf.org/html/rfc8986) (SRv6 Network Programming model).
An overlay based on SRv6 can offer additional features and advantages compared to a traditional IPinIP overlay.

AloysAugustin:
Hi @zvfvrv ! 
Thanks for opening this issue, as discussed offline, we will use gobgp to program VPP with SRv6 routes announced over BGP to achieve this.
In order to do so, we will need to upgrade gobgp (https://github.com/projectcalico/vpp-dataplane/blob/master/go.mod#L35) to a version that supports SRv6 routes, and then update the glue code (https://github.com/projectcalico/vpp-dataplane/blob/master/calico-vpp-agent/routing/bgp_watcher.go#L54) to program the corresponding routes in vpp.

zvfvrv:
Hi @AloysAugustin,
Thanks for your feedback, 
regarding the gobgp upgrade (to latest v2.25.0) can I directly open a PR for that ?

AloysAugustin:
Yes please :)

jjgraham:
Now that https://github.com/projectcalico/vpp-dataplane/pull/186 has merged what is left to do for "Adding support to SRv6 #141" ?

sknat:
Hi @jjgraham @zvfvrv 
I guess we can close this issue now that #186 is merged.
Feel free to re-open another issue if need be.

PR #186, opened 2021-06-01
Support SRv6 

zvfvrv:
This PR is related to the Issue #141


AloysAugustin:
Thanks @zvfvrv , this looks good :)
From a first look I was curious about the inferLocalSidAddr function, this looks like a bit of a hack for now, could you give a bit more info about potential ways to make this work in any cluster?

zvfvrv:
Hi @AloysAugustin, yes inferLocalSidAdd is a dev function that actually I will remove soon :) . The idea is to provide through the resources IPPool and ConfigMap the collection of IP addresses to be used by the agent, in this way it should be completely configurable and should it work in any cluster.

jjgraham:
When do you think SRv6 support will be available in your containers ???

jjgraham:
So close !

calico-vpp-agent/Makefile
```diff
@@ -26,6 +26,7 @@ image: build gobgp
 	@if [ "${ALSO_LATEST}" = "y" ]; then \
 		docker tag calicovpp/agent:$(TAG) calicovpp/agent:latest; \
 	fi
+	docker image prune -f
```
sknat, line 29: Does this have to be part of the `make image` ? 
Because it might impact non related images present on the machine

sknat:
Appart from a minor comment, lgtm 😄

There might be incoming conflicts with #283 but as this PR has been open for long, we can merge it first, and rebase #283 on top.

calico-vpp-agent/Makefile
zvfvrv: Yes I can remove it

sknat: thanks 😄

sknat (approved):

PR #187, opened 2021-06-02
add uplink driver vmxnet3

hedibouattour:


vpplink/vmxnet3.go
```diff
@@ -0,0 +1,94 @@
...
+	"encoding/binary"
+	"fmt"
+	"regexp"
+	"strconv"
+
+	"github.com/pkg/errors"
+	"github.com/projectcalico/vpp-dataplane/vpplink/binapi/vppapi/vmxnet3"
+)
+
+type Vmxnet3Info struct {
```
AloysAugustin, line 28: It would be a bit more consistent to declare a Vmxnet3Interface struct in the `types` package.

AloysAugustin (changes requested):
Looks great! Just a couple comments.

AloysAugustin (approved):

PR #188, opened 2021-06-04
add RescanState for vxlan

hedibouattour:


yaml/base/calico-vpp-daemonset.yaml
```diff
@@ -370,3 +370,12 @@ spec:
         - name: felix-plugins
           hostPath:
             path: /var/lib/calico/felix-plugins
+#---
```
AloysAugustin, line 373: Please remove these comments

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -53,7 +53,25 @@ func (p *VXLanProvider) configureVXLANNodes() error {
...
+	if err != nil {
+		p.log.Errorf("Error listing VXLan tunnels: %v", err)
+	}
+	nodeIP4 := p.server.GetNodeIP(false)
+	nodeIP6 := p.server.GetNodeIP(true)
+	for _, tunnel := range tunnels {
+		if tunnel.SrcAddress.Equal(nodeIP4) || tunnel.SrcAddress.Equal(nodeIP6) {
+			p.log.Infof("Found existing tunnel: %s", tunnel)
+
+			swIfIndex, err := p.vpp.AddVXLanTunnel(tunnel)
```
AloysAugustin, line 68: I think this recreates a tunnel, instead we should use the swIfIndex of the existing tunnel and store it in p.vxlanIfs

AloysAugustin (changes requested):

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -53,7 +53,21 @@ func (p *VXLanProvider) configureVXLANNodes() error {
...
+	p.log.Infof("Rescanning existing VXLAN tunnels")
+	p.vxlanIfs = make(map[string]uint32)
+	tunnels, err := p.vpp.ListVXLanTunnels()
+	if err != nil {
+		p.log.Errorf("Error listing VXLan tunnels: %v", err)
+	}
+	nodeIP4 := p.server.GetNodeIP(false)
+	nodeIP6 := p.server.GetNodeIP(true)
+	for _, tunnel := range tunnels {
+		if tunnel.SrcAddress.Equal(nodeIP4) || tunnel.SrcAddress.Equal(nodeIP6) {
```
AloysAugustin, line 65: Looking at this again, I think we need an additional check here on the VNI, and the ports as well. Other controllers (such as NSM) may create VXLAN tunnels in VPP with the same source IP but a different port / VNI, and we must not use these.

AloysAugustin (changes requested):
One small thing to improve, otherwise this looks good to me :)

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -76,18 +91,20 @@ func (p *VXLanProvider) getVXLANVNI() uint32 {
 	return uint32(*felixConf.VXLANVNI)
 }
 
-func (p *VXLanProvider) AddConnectivity(cn *common.NodeConnectivity) error {
-	p.log.Debugf("Adding vxlan Tunnel to VPP")
-	nodeIP := p.GetNodeIP(vpplink.IsIP6(cn.NextHop))
+func (p *VXLanProvider) getVXLANPORT() uint16 {
```
AloysAugustin, line 94: getVXLANPort since port isn't an acronym :)

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -53,7 +53,22 @@ func (p *VXLanProvider) configureVXLANNodes() error {
...
+	p.vxlanIfs = make(map[string]uint32)
+	tunnels, err := p.vpp.ListVXLanTunnels()
+	if err != nil {
+		p.log.Errorf("Error listing VXLan tunnels: %v", err)
+	}
+	nodeIP4 := p.server.GetNodeIP(false)
+	nodeIP6 := p.server.GetNodeIP(true)
+	for _, tunnel := range tunnels {
+		if (tunnel.SrcAddress.Equal(nodeIP4) || tunnel.SrcAddress.Equal(nodeIP6)) &&
+			tunnel.Vni == p.getVXLANVNI() && tunnel.DstPort == p.getVXLANPORT() && tunnel.SrcPort == p.getVXLANPORT() {
```
AloysAugustin, line 66: It looks like ListVXLanTunnels does not set SrcPort and DstPort on the tunnel struct, so this check is likely to fail (although it may succeed by chance if getVXLANPort returns 0 😉 ).
Also, I wonder what happens when no port is configured: getVXLANPort will return 0, but will the VPP API return 0 or the default port (4789) ?

AloysAugustin (changes requested):
Couple more things to address before we merge this

sknat (approved):

AloysAugustin (approved):

PR #190, opened 2021-06-08
Fix tap0 hanging on tx to linux

AloysAugustin:


AloysAugustin:
The GRO patch updates fix an issue where ipv6 packets could be truncated in the gro code, leading to (very) small packets being sent through tap0, causing the vhost kernel thread to hang. This pr also includes a couple fixes / improvements to the vagrant env & build system.

Makefile
```diff
@@ -86,6 +86,7 @@ restart-calicovpp:
 	kubectl -n calico-vpp-dataplane rollout status ds/calico-vpp-node
 
 .PHONY: goapi
+export VPP_DIR ?= $(shell pwd)/vpp-manager/vpp_build
```
sknat, line 89: Good catch 👍

sknat (approved):

PR #192, opened 2021-06-10
Feature/cleanup tunnel on node drain

hedibouattour:


calico-vpp-agent/routing/connectivity/ipip.go
```diff
@@ -26,11 +26,12 @@ import (
 
 type IpipProvider struct {
 	*ConnectivityProviderData
-	ipipIfs map[string]*types.IPIPTunnel
+	ipipIfs    map[string]*types.IPIPTunnel
+	ipipRoutes map[uint32][]*types.Route
```
sknat, line 30: We should be able to keep only a refcount of routes per tunnel ?
That would simplify the `DelConnectivity` logic

Also we should clear this state when VPP restarts (in `OnVppRestart()`)
And rescan the current state in vpp when we (the agent) restarts `RescanState()`

calico-vpp-agent/routing/connectivity/ipip.go
```diff
@@ -144,18 +152,32 @@ func (p *IpipProvider) DelConnectivity(cn *common.NodeConnectivity) error {
 		return errors.Errorf("Deleting unknown ipip tunnel %s", cn.NextHop.String())
 	}
 	p.log.Infof("IPIP: Del ?->%s %d", cn.NextHop.String(), tunnel.SwIfIndex)
-	err := p.vpp.RouteDel(&types.Route{
+	route_to_delete := &types.Route{
```
sknat, line 155: Ahhh, a snake ! 😄

calico-vpp-agent/routing/connectivity/ipsec.go
```diff
@@ -346,12 +354,34 @@ func (p *IpsecProvider) DelConnectivity(cn *common.NodeConnectivity) (err error)
 		return errors.Errorf("Deleting unknown ipip tunnel %s", cn.NextHop.String())
 	}
 	p.log.Infof("IPSEC: DEL %s via %s [%v]", cn.Dst.String(), cn.NextHop.String(), tunnels)
-	err = p.vpp.RouteDel(&types.Route{
+	route_to_delete := &types.Route{
```
sknat, line 357: Another snake 😄

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -76,17 +92,20 @@ func (p *VXLanProvider) getVXLANVNI() uint32 {
 	return uint32(*felixConf.VXLANVNI)
 }
 
-func (p *VXLanProvider) AddConnectivity(cn *common.NodeConnectivity) error {
-	p.log.Debugf("Adding vxlan Tunnel to VPP")
-	nodeIP := p.GetNodeIP(vpplink.IsIP6(cn.NextHop))
+func (p *VXLanProvider) getVXLANPORT() uint16 {
```
sknat, line 95: `Port` ?

calico-vpp-agent/routing/connectivity/vxlan.go
```diff
@@ -27,12 +27,13 @@ import (
 type VXLanProvider struct {
 	*ConnectivityProviderData
 	vxlanIfs     map[string]uint32
+	vxlanRoutes  map[uint32][]*types.Route
```
sknat, line 30: Same thing

yaml/base/calico-vpp-daemonset.yaml
```diff
@@ -369,4 +369,4 @@ spec:
             path: /run/netns
         - name: felix-plugins
           hostPath:
-            path: /var/lib/calico/felix-plugins
+            path: /var/lib/calico/felix-plugins
```
sknat, line 372: No idea if yaml requires a trailing newline, might be better keeping it...

vpplink/vxlan.go
```diff
@@ -23,6 +23,34 @@ import (
 	"github.com/projectcalico/vpp-dataplane/vpplink/types"
 )
 
+func (v *VppLink) ListVXLanTunnels() ([]*types.VXLanTunnel, error) {
```
sknat, line 26: Might be better to return `[]types.VXLanTunnel` instead of a list of pointers. 
Although I'm guilty of making it wrong in ListIPIP & cie

vpplink/types/vxlan.go
```diff
@@ -26,4 +28,5 @@ type VXLanTunnel struct {
 	DstPort        uint16
 	Vni            uint32
 	DecapNextIndex uint32
+	SwIfIndex      interface_types.InterfaceIndex
```
sknat, line 31: Nice, we could also patch `AddVXLanTunnel` so that it sets this value before returning

sknat (changes requested):
Thanks for the PR !
A few minor remarks (camel vs snake case mostly)
On the route tracking part, we might be missing the re-scanning part (in case the agent restarts), and clearing on VPP restart.

Another thought : we're not handling error cases in `connectivity/*.go` (ipsec has the most visible issues), e.g. the tunnel gets created but we fail to add the route. But we might want to fix this in another PR...

calico-vpp-agent/routing/connectivity/ipip.go
sknat: Re-thinking about this, maybe it's worth keeping a map of the routes (just in case BGP calls us twice with the same route).
We can add the route to VPP only if it's not already present in the map
This will guarantee that on restart we keep the map intact.

So something like `map[uint32]map[string]bool`

calico-vpp-agent/routing/connectivity/ipip.go
```diff
@@ -57,6 +59,26 @@ func (p *IpipProvider) RescanState() {
...
+	}
+	for _, route := range routes {
+		for _, routePath := range route.Paths{
+			for _, tunnel := range p.ipipIfs{
+				if routePath.SwIfIndex == tunnel.SwIfIndex{
+					_, found := p.ipipRoutes[tunnel.SwIfIndex]
+					if !found {
+						p.ipipRoutes[tunnel.SwIfIndex] = make(map[string]bool)
+					}
+					p.ipipRoutes[tunnel.SwIfIndex][route.String()] = true
```
sknat, line 77: `route.Dst.String()` rather no ?

calico-vpp-agent/routing/connectivity/ipip.go
```diff
@@ -57,6 +59,26 @@ func (p *IpipProvider) RescanState() {
...
+
+	p.log.Infof("Rescanning existing routes")
+	p.ipipRoutes = make(map[uint32]map[string]bool)
+	routes, err := p.vpp.GetRoutes(0, false)
+	if err != nil {
+		p.log.Errorf("Error listing routes: %v", err)
+	}
+	for _, route := range routes {
+		for _, routePath := range route.Paths{
+			for _, tunnel := range p.ipipIfs{
```
sknat, line 71: Might be worth adding a temporary `map[uint32]Tunnel` that we would fill in the loop above 
to avoid the nested loop.

sknat (commented):
Just minor comments, other than that, looks good !

calico-vpp-agent/routing/connectivity/ipip.go
hedibouattour: I was hesitating between both, I think it has no impact as long as we keep one format for the key (route.String() or route.Dst.String() )

AloysAugustin (approved):

PR #193, opened 2021-06-11
cloud-init/userdata

onong:


vpp-manager/images/init-eks/init_eks.sh
```diff
@@ -88,11 +113,14 @@ configure_machine ()
 		build_and_install_igb_uio
 		sudo insmod /lib/modules/$(uname -r)/kernel/drivers/uio/igb_uio.ko wc_activate=1
 	fi
+
+	# configure hugepages and persist the config across reboots
 	sudo sysctl -w vm.nr_hugepages=${HUGEPAGES}
 	if [ -f /sys/fs/cgroup/hugetlb/kubepods/hugetlb.2MB.limit_in_bytes ]; then
 		echo $((HUGEPAGES * 2 * 1024 * 1024)) | tee /sys/fs/cgroup/hugetlb/kubepods/hugetlb.2MB.limit_in_bytes
 	fi
-	systemctl restart kubelet
```
sknat, line 95: This might be needed if we want to keep the script (to update hugepage count in kubelet) - if it has not been fixed -

vpp-manager/images/init-eks/init_eks.sh
```diff
@@ -88,11 +113,14 @@ configure_machine ()
...
+	echo "vm.nr_hugepages=${HUGEPAGES}" >> /etc/sysctl.conf
```
sknat, line 122: We might not want to append everytime we run - but same, in the event we want to keep the script used

onong: cloud-init will run only once during first boot thats why this is even required :)

onong: of course, there are ways to make cloud-init run every boot but I wouldn't want that

vpp-manager/images/init-eks/init_eks.sh
onong: Good point. We do want to keep the script. It is needed for the cloud-init/userdata wrapper script and it could also be useful as a standalone script for debugging purposes. And yes, kubelet needs to be restarted when/if we run the script standalone.

scripts/create_eks_cluster.sh
```diff
@@ -0,0 +1,261 @@
...
+	echo "ERROR: Missing clusterSecurityGroup. Exiting..."
+	cd $CUR_DIR; rm -rf $TMP_DIR; exit 1
+fi
+VPC_ID=`aws eks describe-cluster --name $CLUSTER_NAME | grep RESOURCESVPCCONFIG | awk '{print $5}'`
+if [ "$VPC_ID" = "" ]; then
+	echo "ERROR: Missing cluster VPC ID. Exiting..."
+	cd $CUR_DIR; rm -rf $TMP_DIR; exit 1
+fi
+
+### Create security group to allow ssh access
```
AloysAugustin, line 150: Is this SSH access necessary for calico-vpp to start, or just convenient to troubleshoot stuff?
We may not want to open user's instances to the world without explicit confirmation - it may be better to have this disabled by default and optionally enabled.

scripts/create_eks_cluster.sh
```diff
@@ -0,0 +1,261 @@
...
...Vyc2lzdCB0aGUgY2hhbmdlcyBhY3Jvc3MgcmVib290cwoJY2F0IDw8IEVPRiA+PiAvZXRjL3JjLmQvcmMubG9jYWwKbW9kcHJvYmUgdmZpby1wY2kKZWNobyAxID4gL3N5cy9tb2R1bGUvdmZpby9wYXJhbWV0ZXJzL2VuYWJsZV91bnNhZmVfbm9pb21tdV9tb2RlCkVPRgoJY2htb2QgK3ggL2V0Yy9yYy5kL3JjLmxvY2FsCgoJcm0gLXJmICRCVUlMRF9ESVIKfQoKY29uZmlndXJlX21hY2hpbmUgKCkKewoJc3VkbyBybSAtZiAvZXRjL2NuaS9uZXQuZC8xMC1hd3MuY29uZmxpc3QKCXN1ZG8gbW9kcHJvYmUgdWlvCglpZiBbIHgkKGxzbW9kIHwgYXdrICd7IHByaW50ICQxIH0nIHwgZ3JlcCBpZ2JfdWlvKSA9PSB4IF07IHRoZW4KCQlidWlsZF9hbmRfaW5zdGFsbF9pZ2JfdWlvCgkJc3VkbyBpbnNtb2QgL2xpYi9tb2R1bGVzLyQodW5hbWUgLXIpL2tlcm5lbC9kcml2ZXJzL3Vpby9pZ2JfdWlvLmtvIHdjX2FjdGl2YXRlPTEKCWZpCgoJIyBjb25maWd1cmUgaHVnZXBhZ2VzIGFuZCBwZXJzaXN0IHRoZSBjb25maWcgYWNyb3NzIHJlYm9vdHMKCXN1ZG8gc3lzY3RsIC13IHZtLm5yX2h1Z2VwYWdlcz0ke0hVR0VQQUdFU30KCWlmIFsgLWYgL3N5cy9mcy9jZ3JvdXAvaHVnZXRsYi9rdWJlcG9kcy9odWdldGxiLjJNQi5saW1pdF9pbl9ieXRlcyBdOyB0aGVuCgkJZWNobyAkKChIVUdFUEFHRVMgKiAyICogMTAyNCAqIDEwMjQpKSB8IHRlZSAvc3lzL2ZzL2Nncm91cC9odWdldGxiL2t1YmVwb2RzL2h1Z2V0bGIuMk1CLmxpbWl0X2luX2J5dGVzCglmaQoJZWNobyAidm0ubnJfaHVnZXBhZ2VzPSR7SFVHRVBBR0VTfSIgPj4gL2V0Yy9zeXNjdGwuY29uZgp9Cgpjb25maWd1cmVfbWFjaGluZQpjb25maWd1cmVfZHBka19pbnRlcnJ1cHRfbW9kZV9zdXBwb3J0CgotLT09TVlCT1VOREFSWT09LS0K",
```
AloysAugustin, line 164: This blob is pretty scary. Could we re-generate it from source at runtime? Or at least we should add a comment indicating to users how this is generated, so that they can verify the contents if they want.

AloysAugustin (commented):
Couple comments on the auto-EKS script. Love the fact that you generate a cleanup script!

scripts/create_eks_cluster.sh
```diff
@@ -0,0 +1,261 @@
...
+
+### Delete aws-node
+echo; echo
+echo "Deleting aws-node..."
+kubectl delete daemonset -n kube-system aws-node
+
+### Grab the clusterSecurityGroup and VpcId
+CLUSTER_SECURITY_GROUP_ID=`aws eks describe-cluster --name $CLUSTER_NAME | grep RESOURCESVPCCONFIG | awk '{print $2}'`
+if [ "$CLUSTER_SECURITY_GROUP_ID" = "" ]; then
+	echo "ERROR: Missing clusterSecurityGroup. Exiting..."
```
AloysAugustin, line 141: Just tested the script, we should force the output format for the aws CLIs, otherwise this section can fail (on my system aws outputs json by default, which breaks the grep / awk).

scripts/create_eks_cluster.sh
onong: Good point. Yes, ssh access is probably more for dev/test and not end-users. Made it optional.

scripts/create_eks_cluster.sh
onong: Agree about the scary part. Script now generates the stuff at runtime and also added comments explaining what the script does and why.

scripts/create_eks_cluster.sh
```diff
@@ -0,0 +1,369 @@
...
+else
+        echo
+        echo "ERROR: OS_TYPE=$OS_TYPE not supported"
+        exit 1
+fi
+echo "BASE64_COMMAND=$BASE64_COMMAND"
+
+
+### Make sure we have aws cli version 2
+which aws
```
AloysAugustin, line 102: Could be nice to redirect the output of the three `which` commands to `/dev/null` to keep the script output a bit cleaner as well :)

scripts/create_eks_cluster.sh
onong: Fixed!

scripts/create_eks_cluster.sh
onong: On the contrary, I want to see them :)

scripts/create_eks_cluster.sh
onong: I always specify "text" when running "aws configure" so never saw this issue. Thank you for catching this and for the fix :)

AloysAugustin (approved):

PR #199, opened 2021-06-24
Add HostPort support

hedibouattour:


AloysAugustin (changes requested):
Looks good but I think the cleanup part is missing in `DelVppInterface`.

AloysAugustin (commented):
Looks good to me with the latest changes, @sknat could you review as well for the interaction with the cnat?

calico-vpp-agent/cni/network_vpp.go
```diff
@@ -193,6 +193,10 @@ func (s *Server) AddVppInterface(podSpec *storage.LocalPodSpec, doHostSideConf b
 	// Select the first 11 characters of the containerID for the host veth.
 	tunTag := podSpec.NetnsName + "-" + podSpec.InterfaceName
 
+	for idx := range podSpec.HostPorts {
+		s.BindHostPort(&podSpec.HostPorts[idx], podSpec.ContainerIps[0].IP)
```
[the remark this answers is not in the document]
sknat: We should also choose the first `podSpec.ContainerIps` that matches the ip version of `port.HostIP` in order to support dualstack (aka ipv4 + ipv6)

calico-vpp-agent/cni/cni_server.go
```diff
@@ -67,6 +68,15 @@ func NewLocalPodSpecFromAdd(request *pb.AddRequest) (*storage.LocalPodSpec, erro
 		OrchestratorID: request.Workload.Orchestrator,
 		WorkloadID:     request.Workload.Namespace + "/" + request.Workload.Pod,
 		EndpointID:     request.Workload.Endpoint,
+		HostPorts:      make([]storage.HostPortBinding, 0),
+	}
+	for _, port := range request.Workload.Ports {
+		podSpec.HostPorts = append(podSpec.HostPorts, storage.HostPortBinding{
+			HostPort:      port.HostPort,
+			HostIP:        port.HostIp,
```
sknat, line 76: Maybe `net.ParseIP` here, to directly store the parsed `net.IP` ?

calico-vpp-agent/cni/network_vpp.go
```diff
@@ -193,6 +193,18 @@ func (s *Server) AddVppInterface(podSpec *storage.LocalPodSpec, doHostSideConf b
 	// Select the first 11 characters of the containerID for the host veth.
 	tunTag := podSpec.NetnsName + "-" + podSpec.InterfaceName
 
+	for idx := range podSpec.HostPorts {
```
sknat, line 196: Here you might exceed the length of the table, and also we want to expose the ports on all the container addresses. 
Maybe you could use a dedicated function returning the first address that matches 
````golang
for _, hostPort := range podSpec.HostPorts {
   for _, containerAddr := range podSpec.ContainerIps {
       if addrFamilyDiffers (containerAddr, hostPort.HostIP) == nil {
          continue
       }
       err = s.BindHostPort(hostPort, containerAddr)
       ...
   } 
}
````
For the error cleanup, we might want to have `s.BindHostPort` return the EntryID and do `hostPorts.EntryID = id` here. That way if one of the call fails, we can call delete on all the previously created entries.

File: calico-vpp-agent/cni/cni_server.go [hunk 4]
```diff
@@ -98,6 +108,38 @@ func NewLocalPodSpecFromDel(request *pb.DelRequest) *storage.LocalPodSpec {
...
+			},
+		},
+		IsRealIP: true,
+		Proto:    types.TCP,
+		LbType:   types.DefaultLB,
+	}
+	s.log.Infof("(add) %s", entry.String())
+	id, err := s.vpp.CnatTranslateAdd(entry)
+	if err != nil {
+		s.log.Errorf("Error re-injecting cnat entry %s : %v", entry.String(), err)
```
sknat, line 136: We could `return error.Wrapf(err, "...")`

calico-vpp-agent/cni/storage/storage.go
```diff
@@ -124,6 +124,18 @@ type LocalPodSpec struct {
...
+
+	//HostPort/HostIP
+	HostPortsSize int `struc:"int16,sizeof=HostPorts"`
+	HostPorts     []HostPortBinding
+}
+
+type HostPortBinding struct {
+	HostPort      uint32
+	HostIPSize    int `struc:"int16,sizeof=HostIP"`
+	HostIP        net.IP
```
sknat, line 136: Maybe ```` `struc:"[16]byte"` ```` to avoid having `HostIPSize`

AloysAugustin:
Holding merge until the matching libcalico-go and cni-plugin PRs are merged so that we can refer to the upstream .proto in calico-vpp-agent/proto/Makefile .

AloysAugustin (approved):

sknat (approved):

PR #206, opened 2021-07-23
add installation examples for operator

hedibouattour:


AloysAugustin:
Maybe remove the comments at the top of each file since they are related to the operator repo, but otherwise looks good to me. We should merge this once the operator PR is merged upstream.

PR #207, opened 2021-07-28
[WIP] Feature/vcl

sknat:
The resulting network looks as follows. We should support tun only, tun+memif, memif+tun & tun+VCL
loop interfaces carry the pod Address, they are down (grey) for tun+memif/memif+tun & tun only
Other interfaces (memif/tun) inside the VRF are unnumbered on the loopback
 
Networking model


sknat:
Superseeded by #227 
Fixes / updates for k8s 1.22.1 were extracted to #223 #224 #225

PR #208, opened 2021-07-29
dpdk-21.05 ena interrupt support

onong:
Cherry-pick the dpdk-21.05 ena interrupt support patch which is based on the
following patch:

https://patches.dpdk.org/project/dpdk/patch/20210713154118.32111-5-mk@semihalf.com/


AloysAugustin (approved):
Lgtm :)

PR #211, opened 2021-08-09
supporting multiple interfaces

hedibouattour:


vpp-manager/vpp_runner.go
```diff
@@ -631,55 +659,59 @@ func (v *VppRunner) runVpp() (err error) {
...
 
-	err = v.uplinkDriver.CreateMainVppInterface(vpp, vppProcess.Pid)
-	if err != nil {
-		terminateVpp("Error creating main interface (SIGINT %d): %v", vppProcess.Pid, err)
-		v.vpp.Close()
-		<-vppDeadChan
-		return errors.Wrap(err, "Error creating main interface")
-	}
+	for idx := 0; idx < len(v.params.InterfacesSpecs); idx++ {
+		vppifindex, err := v.uplinkDriver[idx].CreateMainVppInterface(vpp, vppProcess.Pid)
```
AloysAugustin, line 663: case

vpp-manager/vpp_runner.go
```diff
@@ -289,41 +315,41 @@ func (v *VppRunner) addExtraAddresses(addrList []netlink.Addr, extraAddrCount in
...
 		if err != nil {
 			log.Errorf("Error adding address to data interface: %v", err)
 		}
 	}
 	return nil
 }
 
-func (v *VppRunner) configureVpp() (err error) {
+func (v *VppRunner) configureVpp(idx int) (err error) {
 	// Always enable GSO feature on data interface, only a tiny negative effect on perf if GSO is not
```
AloysAugustin, line 327: I think it's worth defining variables for the interface spec and config here rather than using the index everywhere.

vpp-manager/vpp_runner.go
```diff
@@ -376,7 +402,7 @@ func (v *VppRunner) configureVpp() (err error) {
 	}
 
 	if v.params.ExtraAddrCount > 0 {
-		err = v.addExtraAddresses(v.conf.Addresses, v.params.ExtraAddrCount)
+		err = v.addExtraAddresses(v.conf.InterfacesConfigs[idx].Addresses, v.params.ExtraAddrCount, v.params.InterfacesSpecs[idx].SwIfIndex)
```
AloysAugustin, line 405: This one we want to do only on the first (main) interface

vpp-manager/startup/startup.go
```diff
@@ -227,9 +235,17 @@ func parseEnvVariables(params *config.VppManagerParams) (err error) {
...
-	params.NativeDriver = ""
-	if conf := getEnvValue(NativeDriverEnvVar); conf != "" {
-		params.NativeDriver = strings.ToLower(conf)
+	nativeDrivers := strings.Split(getEnvValue(NativeDriverEnvVar), "-")
+	for i, driver := range nativeDrivers {
+		if driver != "" {
+			nativeDrivers[i] = strings.ToLower(driver)
+		}
+	}
+	if len(nativeDrivers) != len(mainInterfaces) {
```
AloysAugustin, line 244: I think we may allow an empty native driver if there is only one interface

vpp-manager/vpp_runner.go
```diff
@@ -53,49 +53,75 @@ type VppRunner struct {
...
-		conf:   conf,
+		conf:   confs,
 	}
 }
 
-func (v *VppRunner) Run(driver uplink.UplinkDriver) error {
-	v.uplinkDriver = driver
-	log.Infof("Running with uplink %s", driver.GetName())
+func (v *VppRunner) GenerateVppConfigExecFile() error {
+	template := config.TemplateScriptReplace(v.params.ConfigExecTemplate, v.params, v.conf)
```
AloysAugustin, line 70: It looks like you went through the trouble of modifying the functions handling the config files in the drivers but aren't calling them anywhere - I think thins function and the next one (GenerateVppConfigFile) are the right place to do so.

AloysAugustin (changes requested):
Good starting point, but there are a few things we can improve.
The main one for me is the configuration system, I think the syntax with list of arguments in each environment variable is probably going to be complex to manage in practice.
I'm thinking it may be better to have a single environment variable for the additional interfaces, containing a json array that you could parse to an array of InterfaceSpec, but open to other suggestions as well.

vpp-manager/startup/interface_config.go
```diff
@@ -30,45 +30,68 @@ import (
...
 		log.Warnf("Could not load config from linux (%v), trying file...", err)
-		conf, err2 := loadInterfaceConfigFromFile(params)
+		conf_file, err2 := loadInterfaceConfigFromFile(params)
 		if err2 != nil {
 			log.Warnf("Could not load saved config: %v", err2)
 			// Return original error
 			return nil, err
 		}
-		log.Infof("Loaded config. Interface marked as down since loading config from linux failed.")
+		// If loaded from file replace non loaded interface configs
```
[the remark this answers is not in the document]
hedibouattour: I only do it if it is not successful, just below (65)

vpp-manager/startup/interface_config.go
```diff
@@ -30,45 +30,68 @@ import (
...
+		for i := range conf.InterfacesConfigs {
```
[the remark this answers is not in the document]
hedibouattour: if the length does not match is that an error + I do not save anything?

sknat (changes requested):

vpp-manager/config/config.go
```diff
@@ -83,22 +88,23 @@ type VppManagerParams struct {
...
+	PromiscOn     bool
+	NumTxQueues   int
+	NumRxQueues   int
+	DoSwapDriver  bool
+	Hasv4         bool
+	Hasv6         bool
+	NodeIP4       string
+	NodeIP6       string
+	Mtu           int
+	MainInterface string
```
AloysAugustin, line 107: InterfaceName may be less confusing here as well

vpp-manager/startup/startup.go
```diff
@@ -106,7 +108,7 @@ const (
 )
 
 func GetVppManagerParams() (params *config.VppManagerParams) {
-	params = &config.VppManagerParams{}
+	params = &config.VppManagerParams{InterfacesSpecs: []config.InterfaceSpec{{IsMain: true /* first interface is main */}}}
```
AloysAugustin, line 111: Since you're overwriting params.InterfaceSpecs at the end, maybe instead of initializing params.InterfacesSpec[0] here, it would be a bit cleaner & more explicit to just define a variable `mainInterfaceSpec` of type `config.InterfaceSpec` here that you would use through the function when parsing the different env vars.

AloysAugustin (approved):

PR #221, opened 2021-09-02
Limit corefile number

hedibouattour:


vpp-manager/main.go
```diff
@@ -133,13 +139,44 @@ func makeNewVPPIndex() {
...
+}
+
 func main() {
 	vppDeadChan = make(chan bool, 1)
 	VPPgotSigCHLD = make(map[int]bool)
 	VPPgotTimeout = make(map[int]bool)
 
 	params := startup.GetVppManagerParams()
 
+	cleanup_corefiles(params.CorePattern)
```
AloysAugustin, line 179: The code looks good to me, I just think it would be better to move this towards the end of startup/startup.go, which also sets the core pattern.
The maxCoreFiles definition should go with that file's constants as well.

AloysAugustin (changes requested):

AloysAugustin (approved):

PR #224, opened 2021-09-06
yaml: Update for k8s 1.22.1 & VCL/memif

sknat:


AloysAugustin:
Is there a replacement for the critical pod annotation with this version of k8s?
Also I think we should keep the image tag to prerelease on master and use latest mostly locally for development... Several users are taking the manifests directly from master and the prerelease tag is more stable.

sknat:
Right, I just did `make yaml` which ended up with latest.
I changed the default to `prerelease` agreed it's better to use latest mostly locally

sknat:
> Is there a replacement for the critical pod annotation with this version of k8s?

Afaik it's been replaced starting k8s 1.16 by
````yaml
priorityClassName: system-node-critical
````
the calico/vpp and calico pods have it

AloysAugustin (approved):

PR #226, opened 2021-09-06
vpp: Bump for VCL & memif

sknat:

This changes the mechanism used for punting traffic,
All traffic is punted to a VRF. The default route in this VRF
is via `vpptap0` which results in the same routing for packets
destined to the node address.

This will be used to enable the VCL & punting to the pod tun interfaces


vpplink/feature.go
```diff
@@ -17,12 +17,19 @@ package vpplink
...
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/projectcalico/vpp-dataplane/vpplink/binapi/vppapi/feature"
 	"github.com/projectcalico/vpp-dataplane/vpplink/binapi/vppapi/interface_types"
 )
 
+const (
```
AloysAugustin, line 27: Shouldn't these constants rather go in the cnat file?

vpplink/interfaces.go
```diff
@@ -95,19 +94,7 @@ func (v *VppLink) SetInterfaceMacAddress(swIfIndex uint32, mac *net.HardwareAddr
...
-	}
-	err = v.SetInterfaceVRFAf(swIfIndex, vrfIndex, true)
-	if err != nil {
-		return err
-	}
-	return nil
-}
-
-func (v *VppLink) SetInterfaceVRFAf(swIfIndex, vrfIndex uint32, isIP6 bool) error {
+func (v *VppLink) SetInterfaceVRF(swIfIndex, vrfIndex uint32, isIP6 bool) error {
```
AloysAugustin, line 97: Why remove the helper that does 2 AFs at once?

vpplink/interfaces.go
```diff
@@ -225,24 +212,87 @@ func (v *VppLink) AddInterfaceAddress(swIfIndex uint32, addr *net.IPNet) error {
...
+
+func (v *VppLink) enableDisableInterfaceIP(swIfIndex uint32, isIP6 bool, isEnable bool) error {
+	v.lock.Lock()
+	defer v.lock.Unlock()
 	response := &vppip.SwInterfaceIP6EnableDisableReply{}
-	return v.ch.SendRequest(request).ReceiveReply(response)
+	request := &vppip.SwInterfaceIP6EnableDisable{
+		Enable:    isEnable,
+		SwIfIndex: interface_types.InterfaceIndex(swIfIndex),
+		// Af: types.GetBoolIPFamily(isIP6),
```
AloysAugustin, line 247: This looks weird, what is the behavior if no Af is specified? That's worth a comment at least imo :)

vpplink/memif.go
```diff
@@ -0,0 +1,96 @@
...
+		RxQueues:   uint8(mif.NumRxQueues),
+		TxQueues:   uint8(mif.NumTxQueues),
+		SocketID:   mif.SocketId,
+		BufferSize: uint16(mif.QueueSize),
+	}
+	if mif.MacAddress != nil {
+		request.HwAddr = types.ToVppMacAddress(&mif.MacAddress)
+	}
+	err := v.ch.SendRequest(request).ReceiveReply(response)
+	/* don't defer as memifSocket call also locks */
```
AloysAugustin, line 88: outdated comment?

vpplink/interfaces.go
sknat: I tried with the helpers, doing two AFs at once.
But the issue (generally speaking) comes with error handling.
In the caller you don't know which call generated the error, and if you need to cleanup that gets quickly ugly.

The solution I ended up with is something like :
````go
for _, ipFamily := range vpplink.IpFamilies {
	err = s.vpp.AddDefaultRouteViaTable(..., ipFamily.IsIp6)
	if err != nil {
		return errors.Wrapf(err, "error for %s", ipFamily.Str)
	} else {
		stack.Push(s.vpp.DelDefaultRouteViaTable, ..., ipFamily.IsIp6)
	}
}
````

vpplink/interfaces.go
sknat: Oh, right. Sorry about this,
This is a remainder of trying to move away from unnumbered interfaces, and using 
`SwInterfaceEnableDisable { Af : isIP6 }` but finally reverted to `SwInterfaceIP6EnableDisable`

vpplink/memif.go
sknat: Right...

vpplink/feature.go
sknat: Agreed !

vpplink/interfaces.go
AloysAugustin: Makes sense :)

vpplink/interfaces.go
AloysAugustin: Ah right, missed the IP6 in the message name 😄

AloysAugustin (approved):

PR #229, opened 2021-09-08
Feature/prometheus stats export

hedibouattour:


AloysAugustin:
The way the prometheus library works leads to code that's a bit awkward, it doesn't look super well suited to our use case. I'll try to see if I can find better ones.

vpplink/stats.go
```diff
@@ -0,0 +1,40 @@
...
+
+package vpplink
+
+import (
+	"git.fd.io/govpp.git/adapter"
+	"git.fd.io/govpp.git/adapter/statsclient"
+	"github.com/pkg/errors"
+)
+
+func StatsClientFunc(sc *statsclient.StatsClient) (ifNames adapter.NameStat, dumpStats []adapter.StatEntry, err error) {
```
AloysAugustin, line 24: This should be renamed to something more explicit like GetInterfaceStats

calico-vpp-agent/prometheus/prometheus.go
```diff
@@ -0,0 +1,200 @@
...
+							},
+							Timeseries: []*metricspb.TimeSeries{},
+						}
+						s.lock.Lock()
+						if sta.Type == adapter.SimpleCounterVector {
+							values := sta.Data.(adapter.SimpleCounterStat)
+							for worker := range values {
+								for ifIdx := range values[worker] {
+									if string(ifNames[ifIdx]) != "" {
+										_, swifindex := s.vpp.SearchInterfaceWithName(string(ifNames[ifIdx]))
```
AloysAugustin, line 102: It's not going to be acceptable to do this many API calls to VPP just to gather stats, it would have too much impact on the dataplane, especially since SearchInterfaceWithName is not optimized.
I think ifIdx should already be the sw_if_index, so there should be no need to query vpp.

AloysAugustin (approved):
That's clean, thanks!

PR #230, opened 2021-09-08
agent: support BGPConfigurationOptions

hedibouattour:


AloysAugustin (changes requested):

calico-vpp-agent/routing/routing_server.go
```diff
@@ -214,26 +240,41 @@ func (s *Server) getDefaultBGPConfig(log *logrus.Entry, clientv3 calicov3cli.Int
...
-			LogSeverityScreen:     "INFO",
+			LogSeverityScreen:     "Info",
 			NodeToNodeMeshEnabled: &b,
 		}
 		asn, err := numorstring.ASNumberFromString("64512")
 		if err != nil {
 			return nil, err
 		}
 		ret.ASNumber = &asn
+		ret.ListenPort = 179
```
AloysAugustin, line 263: This should go in the struct initialization (l.255), asn is out only because we need to call ASNumberFromString

calico-vpp-agent/routing/routing_server.go
```diff
@@ -199,6 +205,26 @@ func (s *Server) serveOne() error {
...
 }
 
+func (s *Server) getBGPConf(log *logrus.Entry, clientv3 calicov3cli.Interface) (*calicov3.BGPConfigurationSpec, error) {
+	defaultBGPConf, err := s.getDefaultBGPConfig(log, clientv3)
+	if err != nil {
+		return nil, errors.Wrap(err, "error getting default BGP configuration")
+	}
+	nodeSpecificConf, err := clientv3.BGPConfigurations().Get(context.Background(), "node."+config.NodeName, options.GetOptions{})
+	if err == nil {
+		defaultBGPConf.ListenPort = nodeSpecificConf.Spec.ListenPort
```
AloysAugustin, line 215: We should verify that these are set in the node-specific config before overriding the default value.
Also it's a bit more idiomatic go to arrange the code like this:
```
if (err != nil) {
    ... error handling
}
... success case
```

AloysAugustin (changes requested):
This is looking good, just a few minor things to adjust

AloysAugustin (approved):

PR #250, opened 2021-10-14
agent: fix bgppeers handling

hedibouattour:


calico-vpp-agent/routing/watchers/peers_watcher.go
```diff
@@ -118,43 +193,61 @@ func (w *PeerWatcher) WatchBGPPeers() error {
 				}
 				ip := peer.Spec.PeerIP
 				asn := uint32(peer.Spec.ASNumber)
```
AloysAugustin, line 195: Same comment as before

AloysAugustin (commented):
This is a good start but there is a missing part, nodes can be added and removed during the life of the cluster, and when they do we need to add / remove BGP peerings depending on the peerSelectors that apply to us. Would you prefer to merge this PR and add that in a separate PR or to do everything in this PR?

AloysAugustin:
Rebased on master

AloysAugustin (approved):

PR #259, opened 2021-10-22
agent: add host policies

hedibouattour:


calico-vpp-agent/policy/policy_server.go
```diff
@@ -579,12 +674,79 @@ func (s *Server) handleActiveProfileRemove(msg *proto.ActiveProfileRemove, pendi
...
+		for _, swifindex := range hep.UplinkSwIfIndexes {
+			hep.TapSwIfIndexes = append(hep.TapSwIfIndexes, s.uplinkToTap[swifindex])
+		}
+	} else if hep.InterfaceName == "*" {
+		for uplink, tap := range s.uplinkToTap {
+			hep.UplinkSwIfIndexes = append(hep.UplinkSwIfIndexes, uplink)
+			hep.TapSwIfIndexes = append(hep.TapSwIfIndexes, tap)
+		}
+	}
+	for _, provider := range s.routingServer.ConnectivityServer.Providers {
```
AloysAugustin, line 704: This isn't great from an encapsulation perspective, the policy server shouldn't need to know that there are several Providers and a Connectivity server in the routing server. I think it would be a bit better to add a GetAllTunnels method on the routing server. This way, we could switch the ConnectivityServer and the Providers back to private (lowercase) as well.

calico-vpp-agent/policy/policy_server.go
```diff
@@ -579,12 +674,79 @@ func (s *Server) handleActiveProfileRemove(msg *proto.ActiveProfileRemove, pendi
...
+		hep.TunnelSwIfIndexes = append(hep.TunnelSwIfIndexes, provider.GetSwifindexes()...)
+	}
+	if len(hep.UplinkSwIfIndexes) == 0 || len(hep.TapSwIfIndexes) == 0 {
+		return errors.Errorf("No interface to configure as host endpoint")
```
AloysAugustin, line 708: We should specify the name of the host endpoint that triggers this error in the message to ease debugging.

calico-vpp-agent/policy/policy_server.go
```diff
@@ -317,87 +369,126 @@ func (s *Server) Stop() {
...
+					if newTunnel {
+						h.TunnelSwIfIndexes = append(h.TunnelSwIfIndexes, tunnelChange.Swifindex)
+						s.log.Infof("Configuring policies on added tunnel [%d]", tunnelChange.Swifindex)
+						err = s.vpp.ConfigurePolicies(tunnelChange.Swifindex, h.currentForwardConf)
+						if err != nil {
+							s.log.Errorf("cannot configure policies on tunnel interface %d", tunnelChange.Swifindex)
+						}
+					}
+				} else { // delete case
+					s.log.Infof("Tunnel [%d] is deleted", tunnelChange.Swifindex)
```
AloysAugustin, line 483: We might want to either remove this log or add more info specific to the policy server, as right now I think tunnels add / delete are already logged in the connectivity server.

calico-vpp-agent/policy/policy_server.go
```diff
@@ -317,87 +369,126 @@ func (s *Server) Stop() {
...
+					newTunnel := true
+					for _, v := range h.TunnelSwIfIndexes {
+						if v == tunnelChange.Swifindex {
+							newTunnel = false
+						}
+					}
+					if newTunnel {
+						h.TunnelSwIfIndexes = append(h.TunnelSwIfIndexes, tunnelChange.Swifindex)
+						s.log.Infof("Configuring policies on added tunnel [%d]", tunnelChange.Swifindex)
+						err = s.vpp.ConfigurePolicies(tunnelChange.Swifindex, h.currentForwardConf)
```
AloysAugustin, line 477: We should call ConfigurePolicies only if we aren't `pending`? Unless we have the guarantee that `h.currentForwardConf` always make this call do nothing when `h` hasn't been created (even if, for instance, it was created once and then VPP restarted).

Also, these few lines look like details that are internal to the HostEndpoint logic, and I think it would be better to put them in a method on the HostEndpoint struct.

calico-vpp-agent/policy/policy_server.go
```diff
@@ -317,87 +369,126 @@ func (s *Server) Stop() {
...
+						if err != nil {
+							s.log.Errorf("cannot configure policies on tunnel interface %d", tunnelChange.Swifindex)
+						}
+					}
+				} else { // delete case
+					s.log.Infof("Tunnel [%d] is deleted", tunnelChange.Swifindex)
+					for index, existingSwifindex := range h.TunnelSwIfIndexes {
+						if existingSwifindex == tunnelChange.Swifindex {
+							h.TunnelSwIfIndexes = append(h.TunnelSwIfIndexes[:index], h.TunnelSwIfIndexes[index+1:]...)
```
AloysAugustin, line 486: Might be nice to add a comment saying that we don't delete the policies because they are auto-deleted when interfaces are removed, otherwise this could look like a bug.

AloysAugustin (changes requested):
A few minor comments, otherwise looks very good 👍

calico-vpp-agent/policy/policy_server.go
```diff
@@ -317,87 +369,126 @@ func (s *Server) Stop() {
...
+			if err != nil {
+				s.log.WithError(err).Error("Error processing update from felix, restarting")
+				conn.Close()
+				s.felixRestarted <- true
+				// TODO: Restart VPP as well? State is left over there...
+				return
 			}
+		case tunnelChange := <-s.routingServer.ConnectivityServer.GetTunnelChangeChan():
 			var pending bool
 			if s.state == StateSyncing {
```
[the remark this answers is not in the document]
AloysAugustin: Looks like StateConnected was added to the wrong condition, in Connected we should have pending=true

AloysAugustin (changes requested):
Just one last small fix, otherwise lgtm 🙂

AloysAugustin (approved):

PR #270, opened 2021-11-26
agent: check available buffers before creating an interface

hedibouattour:


calico-vpp-agent/cni/cni_server.go
```diff
@@ -55,6 +58,8 @@ type Server struct {
 	tuntapDriver   *pod_interface.TunTapPodInterfaceDriver
 	vclDriver      *pod_interface.VclPodInterfaceDriver
 	loopbackDriver *pod_interface.LoopbackPodInterfaceDriver
+
+	availableBuffers float64
```
AloysAugustin, line 62: Is there a reason not to use uint64 ? Looks like it could reduce the number of casts

calico-vpp-agent/cni/cni_server.go
```diff
@@ -319,6 +324,18 @@ func NewServer(v *vpplink.VppLink, rs *routing.Server, ps *policy.Server, prs *p
...
 	pb.RegisterCniDataplaneServer(server.grpcServer, server)
 	l.Infof("Server starting")
+
+	sc := statsclient.NewStatsClient("")
+	err = sc.Connect()
+	if err != nil {
+		server.log.WithError(err).Errorf("could not connect statsclient")
+		return nil, err
+	}
+	time.Sleep(time.Second * 15) //waiting for stats to get updated on vpp manager side
```
AloysAugustin, line 334: We need a better way to wait for stats to be updated. The CNI server should not be started before vpp-manager has finished configuring VPP, are the stats not up-to-date at this point?

AloysAugustin (changes requested):

vpplink/stats.go
```diff
@@ -38,3 +41,15 @@ func GetInterfaceStats(sc *statsclient.StatsClient) (ifNames adapter.NameStat, d
 	}
 	return ifNames, dumpStats, nil
 }
+
+func (v *VppLink) GetBufferStats() (uint32, uint32, uint32, error) {
+	response := &interfaces.GetBuffersStatsReply{}
```
AloysAugustin, line 46: This function is missing the `v.lock... defer unlock` boilerplate (and there's also a missing newline at the end of the file).

calico-vpp-agent/cni/network_vpp.go
```diff
@@ -30,6 +30,18 @@ func getInterfaceVrfName(podSpec *storage.LocalPodSpec, suffix string) string {
...
 }
 
+func (s *Server) checkAvailableBuffers() (uint64, error) {
+	NumRxQueues := config.TapNumRxQueues
+	NumTxQueues := config.TapNumTxQueues
+	RxQueueSize := vpplink.DefaultIntTo(config.TapRxQueueSize, vpplink.DEFAULT_QUEUE_SIZE)
+	TxQueueSize := vpplink.DefaultIntTo(config.TapTxQueueSize, vpplink.DEFAULT_QUEUE_SIZE)
+	buffersNeeded := uint64(RxQueueSize * NumRxQueues + TxQueueSize * NumTxQueues)
+	if buffersNeeded > s.availableBuffers {
+		return buffersNeeded, errors.Errorf("Cannot create interface: Out of buffers: available buffers = %d, buffers needed = %d", s.availableBuffers, buffersNeeded)
```
AloysAugustin, line 40: It's great to indicate the values in the message, to make it even better we could tell the user how to increase the number of buffers and how to prevent the error by appending something like: `Increase buffers-per-numa in the VPP configuration or reduce CALICOVPP_TAP_RING_SIZE the to allow more pods to be scheduled. You may want to limit the number of pods per node to prevent this error`.

AloysAugustin (changes requested):

AloysAugustin (approved):
Looks good 👍

PR #276, opened 2021-12-06
agent: support service IPs in BGP configuration

hedibouattour:


AloysAugustin (changes requested):
Looks OK, but the withdrawal of specific routes (when a local service instance is deleted) seems to be missing.

AloysAugustin (approved):

PR #271, opened 2021-12-01
vpp: support multi tx queues

hedibouattour:

PR #279, opened 2021-12-15
Integrate updated memif API

sknat:
This updates the API for creating memif sockets per changes required for upstream
- memif socket IDs are now allocated opportunistically walking the hash
- netns name is part of the socket name


sknat:
This should superseed #271

AloysAugustin (approved):

PR #282, opened 2021-12-24
agent: handle service IP addresses change, add support for loadbalanc…

hedibouattour:
…erIPs

### DIFF
--- a/calico-vpp-agent/services/service_handler.go
+++ b/calico-vpp-agent/services/service_handler.go
@@ -108,48 +108,30 @@ func (s *Server) updateCnatEntry(key string, entry *types.CnatTranslateEntry) (e
 	return nil
 }
 
-func (s *Server) advertiseSpecificRoute(IPAddress net.IP, withdraw bool) {
-	_, serviceExternalIPs, serviceLoadbalancerIPs := s.getServiceIPs()
-	for _, serviceExternalIP := range serviceExternalIPs {
-		_, ipnet, err := net.ParseCIDR(serviceExternalIP.CIDR)
-		if err != nil {
-			s.log.Error(err)
-		}
-		if ipnet.Contains(IPAddress) {
-			if withdraw {
-				common.SendEvent(common.CalicoVppEvent{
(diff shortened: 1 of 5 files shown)


calico-vpp-agent/services/service_handler.go
```diff
@@ -104,26 +108,47 @@ func (s *Server) updateCnatEntry(key string, entry *types.CnatTranslateEntry) (e
 	return nil
 }
 
-func (s *Server) advertiseSpecificRouteForExternalIP(extIP net.IP, withdraw bool) {
-	_, serviceExternalIPs, _ := s.getServiceIPs()
+func (s *Server) advertiseSpecificRoute(IPAddress net.IP, withdraw bool) {
```
sknat, line 111: It might be worth adding a function that returns whether a list of IPBlocks contains an address or not.
this function could then be simplified as :
````
_, serviceBlocks, serviceLBBlocks := s.getServiceIPs()
if blockContains(serviceBlocks, addr) || blockContains(serviceLBBlocksm addr) {
    common.SendEvent (...)
}
````
Otherwise we're a bit under the impression this is a loop over all externalIPs

File: calico-vpp-agent/services/service_handler.go [hunk 2]
```diff
@@ -171,6 +196,28 @@ func (s *Server) addServicePort(service *v1.Service, ep *v1.Endpoints) (err erro
 			}
 		}
 
+		for _, ingress := range service.Status.LoadBalancer.Ingress {
+			if ingressIP := net.ParseIP(ingress.IP); ingressIP != nil {
+				if entry, err := getCalicoEntry(&servicePort, ep, ingressIP, localOnly); err == nil {
+					if localOnly {
+						if len(entry.Backends) > 0 {
+							s.advertiseSpecificRoute(ingressIP, false)
```
sknat, line 204: Not sure we need to advertise the ingressIP also, do we ?

sknat: My bad, ok we do

sknat (approved):

PR #284, opened 2022-01-04
vpp-manager: Make vpp own netns persistant

sknat:
It appears process termination does delete veth interfaces put in its net namespace - 
which is not true for physical interfaces.
This makes the parking netns used for the uplink interfaces taken by VPP a persistant one, 
created with iproute2.

### DIFF
--- a/vpp-manager/vpp_runner.go
+++ b/vpp-manager/vpp_runner.go
@@ -27,6 +27,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
 	calicoapi "github.com/projectcalico/libcalico-go/lib/apis/v3"
 	calicocli "github.com/projectcalico/libcalico-go/lib/clientv3"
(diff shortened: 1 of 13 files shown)


vpp-manager/vpp_runner.go
```diff
@@ -697,24 +699,38 @@ func (v *VppRunner) pingCalicoVpp() error {
...
+		}
 	}
-	vppCmd.Stdout = os.Stdout
-	vppCmd.Stderr = os.Stderr
-	err = vppCmd.Start()
 
+	/**
+	 * From this point it is very important that every exit
+	 * path calls restoreConfiguration after vpp exits */
 	defer v.restoreConfiguration()
```
AloysAugustin, line 714: Is this necessary before we start vpp now?

sknat: I didn't see any drawbacks to running it even if VPP startup fails,
This allows to restore connectivity in the case VPP manages to grab the
uplink (e.g. dpdk), and still fail to start afterwards (e.g. oom in another plugin init)
Although this can arguably sound farfetched 😄

AloysAugustin (approved):

PR #285, opened 2022-01-05
Operator based installation

AloysAugustin:
- yaml: switch to operator installs
- yaml: regenerate manifests


AloysAugustin:
@zvfvrv we're changing our installation method to use the calico operator for the calico parts. I tried to update the srv6 YAMLs, but please check whether this PR breaks anything if you can.

sknat (approved):

zvfvrv:
Hi @AloysAugustin , I believe further configuration changes need to be made for SRv6,
I tried to create a deployment based on this PR but strangely I couldn't get a working execution,
even if I use the basic version without SRv6. The commands I performed are:
`make start-test-cluster` followed by `make test-install-calicovpp`. Is it necessary to do anything else?

AloysAugustin:
Thanks for giving a try @zvfvrv , you may need to run `make -C test/vagrant box` beforehand as the base image for the node VMs has been updated.
Let me know if it still fails with that.

zvfvrv:
Thanks for the hint, it resolved that problem. 
Regarding SRv6 there is just one thing that currently is not working: the IPv6 assignment to the pods.
I was not able to configure the operator in order to enable the ipv6 assignment with ipam, with the old method I was able to set the parameter "assign_ipv6": "true", from my understanding it should be enough to define an ippool for each ip family.
If I define just the ipv4 pool, all the things work, but the pods doesn't have an ipv6 assigned of course.
If I define just the ipv6 pool or both of them, which should be the goal, the cluster doesn't work properly and the pods like calico-node doesn't come up.
Any further suggestion? Thanks

AloysAugustin:
How did you configure IPv6, did you try to add both an IPv4 pool and an IPv6 pool to the Installation spec for the operator, or did you define the pools separately?

zvfvrv:
Yes with the Installation spec for the operator something like:
....

    ipPools:
      -  cidr: 172.16.0.0/16 
      -  cidr: "fd01::/64"
   
.....

AloysAugustin:
Ah, with IPv6 we've seen an issue where the the operator refuses to setup IPv6 connectivity with encapsulation enabled as the Linux / eBPF dataplanes don't support that. This may be the cause of the problem you're seeing. We're working on a fix and we'll submit a PR to address this in the operator repo.

AloysAugustin:
@zvfvrv Since there's nothing we can do in the short term I think we'll merge this PR. I'll let you know once we submit the fix on the operator. SRv6 will continue to work in the previous release, and we should be able to address this problem before the next one.

PR #312, opened 2022-01-28
events: fix deleted tunnel events

AloysAugustin:


sknat (approved):
Yep... 
Good catch, thanks !

PR #340, opened 2022-02-25
cni: fix buffer accounting

AloysAugustin:
We used not to re-increment the number of available buffers when a tap interface was deleted. This fixes that.


hedibouattour:
so every time the check is done on all existing pods (+ host tap), not only the new added one?

AloysAugustin:
With this we don't modify `availableBuffers` as pods are added / removed anymore, and we just check that the total number of buffers we'd use with one more tap fits in the original number of available buffers.
I think it makes it simpler as we don't need to keep track of all the interfaces add / del to update the available buffer count.

sknat (approved):

PR #355, opened 2022-03-17
agent: better VRF tags

sknat:
This make the VRF tag be a hash (netns, ipversion),
followed by the ip version, and the netns basename truncated to 63 chars
Matching existing VRFs is then done based on this tag.

This allows to keep a user friendly display in the debug CLI
while preventing toe stepping on restarts (where existing VrfIds
get reallocated)

### DIFF
--- a/calico-vpp-agent/cni/storage/storage.go
+++ b/calico-vpp-agent/cni/storage/storage.go
@@ -17,21 +17,27 @@ package storage
 
 import (
 	"bytes"
+	"crypto/sha512"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/lunixbochs/struc"
 	"github.com/pkg/errors"
 	"github.com/projectcalico/vpp-dataplane/calico-vpp-agent/common"
 	"github.com/projectcalico/vpp-dataplane/calico-vpp-agent/config"
+	"github.com/projectcalico/vpp-dataplane/vpplink"
 	"github.com/projectcalico/vpp-dataplane/vpplink/types"
 )
 
 const (
-	CniServerStateFileVersion = 5 // Used to ensure compatibility wen we reload data
+	CniServerStateFileVersion = 5  // Used to ensure compatibility wen we reload data
+	MaxApiTagLen              = 63 /* No more than 64 characters in API tags */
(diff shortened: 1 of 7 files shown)


calico-vpp-agent/cni/storage/storage.go
```diff
@@ -238,8 +244,29 @@ func (hp *HostPortBinding) String() string {
...
+
+func truncateStr(text string, size int) string {
+	if len(text) > size {
+		return text[:size]
+	}
+	return text
+}
+
+func (ps *LocalPodSpec) GetVrfTag(ipFamily vpplink.IpFamily) string {
+	h := hash(fmt.Sprintf("%s%s", ps.NetnsName, ipFamily.ShortStr))
```
AloysAugustin, line 261: Why don't we include the interface name in the hashed string / tag for the VRF? Won't we have identical tags if a pod requests 2 interfaces (multi-net or just plain multus) ?

calico-vpp-agent/cni/storage/storage.go
```diff
@@ -238,8 +244,29 @@ func (hp *HostPortBinding) String() string {
...
+	s := fmt.Sprintf("%s-%s-%s", h, ipFamily.ShortStr, filepath.Base(ps.NetnsName))
+	return truncateStr(s, MaxApiTagLen)
+}
+
 func (ps *LocalPodSpec) GetInterfaceTag(prefix string) string {
-	return fmt.Sprintf("%s-%s-%s", prefix, ps.NetnsName, ps.InterfaceName)
+	h := hash(fmt.Sprintf("%s-%s-%s", prefix, ps.NetnsName, ps.InterfaceName))
+	s := fmt.Sprintf("%s-%s-%s", h, ps.InterfaceName, filepath.Base(ps.NetnsName))
```
AloysAugustin, line 268: Also for the multi-net case, I think it may facilitate troubleshooting if we added the interface name to the tag (before the NetnsName so it doesn't truncated, it would be annoying if it got truncated to just eth ;) )

calico-vpp-agent/cni/storage/storage.go
sknat: Right, fixed :)

calico-vpp-agent/cni/storage/storage.go
sknat: Same

AloysAugustin (approved):

PR #361, opened 2022-03-24
Govpp templates

sknat:
This is an experimentation to push some logic back to govpp
in order to streamline it.

It depends on https://gerrit.fd.io/r/c/govpp/+/33494 to be merged to work


sknat:
Close in favor of #430

PR #369, opened 2022-04-01
agent: improve ipsec, cancel goroutines

sknat:
This patch introduces a cancel call for the goroutine checking
the tunnel's status in case it gets deleted.

It also make the ipsec connectivity leverage the cleanup stack
in case of add failures

### DIFF
--- a/calico-vpp-agent/connectivity/ipsec.go
+++ b/calico-vpp-agent/connectivity/ipsec.go
@@ -29,9 +29,31 @@ import (
 	"github.com/projectcalico/vpp-dataplane/vpplink/types"
 )
 
+type IpsecTunnel struct {
+	*types.IPIPTunnel
+	cancel func()
+}
+
+func NewIpsecTunnel(ipipTunnel *types.IPIPTunnel) *IpsecTunnel {
+	return &IpsecTunnel{IPIPTunnel: ipipTunnel, cancel: func() {}}
+}
+
+func ipToSafeString(addr net.IP) string {
+	return strings.ReplaceAll(strings.ReplaceAll(addr.String(), ".", "_"), ":", "_")
+}
+
+func (tunnel *IpsecTunnel) Profile() string {
+	return fmt.Sprintf("pr_%s_to_%s", ipToSafeString(tunnel.Src), ipToSafeString(tunnel.Dst))
+}
+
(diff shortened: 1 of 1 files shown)


calico-vpp-agent/connectivity/ipsec.go
```diff
@@ -111,211 +133,178 @@ func (p *IpsecProvider) RescanState() {
...
-		}
-
-		if bytes.Compare(tunnel.Src.To4(), tunnel.Dst.To4()) > 0 {
-			p.log.Warnf("IPIP tunnel still down, re-trying initiate IKE for IPsec tunnel=%s", tunnel.String())
-			err = p.vpp.IKEv2Initiate(profile)
+func (p *IpsecProvider) waitForIPsecSA(tunnel IpsecTunnel) func() {
+	alive := true
+	go (func() {
+		for alive {
+			time.Sleep(time.Second)
```
AloysAugustin, line 281: We sleep after checking for alive, so the tunnel may have been deleted before the next API calls.
It might actually be cleaner to do this with a `close` channel inside the IpsecTunnel object instead of a cancel func, then the wait function can be written with a select on that channel and on a ticker, and cancelling the tunnel retries is as simple as closing the channel.

calico-vpp-agent/connectivity/ipsec.go
```diff
@@ -392,19 +404,18 @@ func (p *IpsecProvider) DelConnectivity(cn *common.NodeConnectivity) (err error)
 	}
 	err = p.vpp.RouteDel(routeToDelete)
 	if err != nil {
-		return errors.Wrapf(err, "Error deleting route ipip tunnel %v: %v", tunnels)
+		p.log.Errorf("Error deleting route ipip tunnel %v: %v", tunnels, err)
```
AloysAugustin, line 407: Is it not an issue in vpp to remove the tunnel if there is still a route using it?

sknat: It's debatable, I'd say it's better to still try to remove the tunnel if needed. 
But whatever we do, we might have leftover state at this point.
It's just that in case of a fail due to a missing route it's better to try to run the rest of the delete

AloysAugustin (approved):

PR #371, opened 2022-04-06
agent: better Nodes watcher

sknat:


calico-vpp-agent/watchers/nodes_watcher.go
```diff
@@ -279,55 +214,81 @@ func (w *NodeWatcher) configureRemoteNodeSnat(node *common.NodeState, isAdd bool
...
 		Type: common.PeerNodeStateChanged,
-		Old:  &old.Node,
-		New:  &node.Node,
+		Old:  old,
+		New:  node,
 	})
 
-	if common.GetStringChangeType(oldV4IP, newV4IP) > common.ChangeSame ||
-		common.GetStringChangeType(oldV6IP, newV6IP) > common.ChangeSame {
+	change := common.MaxChangeType(common.GetStringChangeType(oldV4IP, newV4IP), common.GetStringChangeType(oldV6IP, newV6IP))
```
AloysAugustin, line 251: What does the change order mean? I think we should make this easier to understand.

sknat: Agreed, switched to a flag based logic, should be more straightforward than the ordering

AloysAugustin: Way easier to understand, thanks!

AloysAugustin (approved):

PR #381, opened 2022-05-03
Feature/multinet services

hedibouattour:


calico-vpp-agent/watchers/common.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+func NewClient(timeout time.Duration, addToSchemes []func(s *runtime.Scheme) error) (*client.WithWatch, error) {
```
sknat, line 27: Maybe this should be closer to the other k8sclient definitions (in `cmd/calico_vpp_dataplane.go`) also I'm wondering if it's possible to have a single `k8sclient` object (and connection) for both the `serviceServer` and the `net_watcher` but I'm not too familiar with the `NewScheme` logic

calico-vpp-agent/watchers/net_watcher.go
```diff
@@ -0,0 +1,210 @@
...
+			return err
+		}
+		networks[netDef.Name] = netDef
+	}
+	common.SendEvent(common.CalicoVppEvent{
+		Type: common.NetsSynced,
+		New:  networks,
+	})
+
+restart:
```
sknat, line 90: `for {}` ?

calico-vpp-agent/watchers/net_watcher.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	common.SendEvent(common.CalicoVppEvent{
+		Type: common.NetDeleted,
+		Old:  netDef,
+	})
+	return nil
+}
+
+func getNetworkVrfName(networkName string, suffix string) string {
+	return fmt.Sprintf("pod-%s-table-%s", networkName, suffix)
```
sknat, line 157: Just as a note, the name is capped at 64 characters, but I guess this should be enough (I don't know what network names typically are). Otherwise, it's probably worth doing something like  `GetVrfTag`

global-watcher/watcher.go
```diff
@@ -0,0 +1,326 @@
...
+	AddToScheme := SchemeBuilder.AddToScheme
+	SchemeBuilder.Register(&v1.ServiceList{}, &v1.PodList{}, &v1.Pod{}, &v1.Service{}, &v1.Endpoints{})
+
+	k8sClient, err := watchers.NewClient(10*time.Second, []func(s *runtime.Scheme) error{AddToScheme, nadv1.AddToScheme})
+	if err != nil {
+		log.Errorf("failed instantiating kubernetes client: %v", err)
+	}
+	kubernetesClient = *k8sClient
+
+restart:
```
sknat, line 66: same, maybe a `for {}` ?

global-watcher/watcher.go
```diff
@@ -0,0 +1,326 @@
...
+	svcWatcher, err := kubernetesClient.Watch(context.Background(), svcList, &client.ListOptions{})
+	if err != nil {
+		log.Errorf("couldn't watch services: %s", err)
+	}
+	nadWatcher, err := kubernetesClient.Watch(context.Background(), nadList, &client.ListOptions{})
+	if err != nil {
+		log.Errorf("couldn't watch network attachment definitions: %s", err)
+	}
+
+	watching(podWatcher, svcWatcher, nadWatcher)
```
sknat, line 84: Should we handle the return value ?

global-watcher/watcher.go
```diff
@@ -0,0 +1,326 @@
+// Copyright (C) 2019 Cisco Systems Inc.
```
sknat, line 1: 2022 😄

vpp-manager/vpp_runner.go
```diff
@@ -394,7 +394,7 @@ func (v *VppRunner) setupTapVRF(ifSpec *config.InterfaceSpec, ifState *config.Li
 		// default route in default table
 		err = v.vpp.AddDefaultRouteViaTable(vrfId, common.DefaultVRFIndex, ipFamily.IsIp6)
 		if err != nil {
-			return []uint32{}, errors.Wrapf(err, "error adding VRF %d default route via VRF %d", vrfId, common.PodVRFIndex)
+			return []uint32{}, errors.Wrapf(err, "error adding VRF %d default route via VRF %d", vrfId, common.DefaultVRFIndex)
```
sknat, line 397: 👍

global-watcher/version
```diff
@@ -0,0 +1,21 @@
+Image tag                   : latest
```
sknat, line 1: Missing `.gitignore` ?

global-watcher/Makefile
```diff
@@ -0,0 +1,16 @@
...
+ALSO_LATEST ?= n # If 'y' also tag images with :latest
+
+build:
+	go build -o ./watcher .
+
+image: build
+	@echo "Image tag                   : $(TAG)"                         > $(VERSION_FILE)
+	@echo "VPP-dataplane version       : $(shell git log -1 --oneline)" >> $(VERSION_FILE)
+	@cat $(GENERATE_LOG_FILE)                                           >> $(VERSION_FILE)
+	docker build --pull -t calicovpp/global-watcher:$(TAG) .
```
sknat, line 13: Just a naming question, but `global-watcher` might be a bit too generic, what about something like `multinet-ep-monitor` or just `multinet-monitor` ?

global-watcher/watcher.go
```diff
@@ -0,0 +1,326 @@
...
+			if !ok {
+				log.Warn("pods watch channel closed")
+				return true
+			}
+			switch update.Type {
+			case watch.Added:
+				pod := update.Object.(*v1.Pod)
+				podNadIf, inNetwork := pod.Annotations["k8s.v1.cni.cncf.io/networks"]
+				if inNetwork {
+					podNad := podNadIf[0:strings.Index(podNadIf, "@")]
```
sknat, line 132: `strings.Split(podNadIf, "@")[0]` ?

calico-vpp-agent/watchers/net_watcher.go
```diff
@@ -0,0 +1,210 @@
...
+		return err
+	}
+	networks := make(map[string]*NetworkDefinition)
+	for _, net := range netList.Items {
+		netDef, err := w.OnNetAdded(&net)
+		if err != nil {
+			w.log.Error(err)
+			return err
+		}
+		networks[netDef.Name] = netDef
```
sknat, line 83: An potential option in order not to send the whole map over the chan would be to pass the networks sperately, and use the `NetsSynced` event just as a signaling mechanism.
````
for _, net := range netList.Items {
        ...
        // send an event for each added network in the loop
        common.SendEvent(common.CalicoVppEvent{
		Type: common.NetAdded,
		New:  netDef.Copy(),
	})
       ...
}
// And send a NetsSynced event just to signal that everything is in sync
common.SendEvent(common.CalicoVppEvent{Type: common.NetsSynced})
````

File: calico-vpp-agent/cni/cni_server.go [hunk 12]
```diff
@@ -378,9 +405,32 @@ func (s *Server) ServeCNI(t *tomb.Tomb) error {
...
+
+	netsSynced := make(chan bool)
+	nets := make(map[string]*watchers.NetworkDefinition)
+	go func() {
+		for t.Alive() {
+			event := <-s.cniMultinetEventChan
+			switch event.Type {
+			case common.NetsSynced:
+				nets = event.New.(map[string]*watchers.NetworkDefinition)
+				s.networkDefinitions = nets
```
sknat, line 417: And so here we wouldn't need this

calico-vpp-agent/cni/cni_server.go
```diff
@@ -378,9 +405,32 @@ func (s *Server) ServeCNI(t *tomb.Tomb) error {
...
+				netsSynced <- true
+			case common.NetAdded:
+				netDef := event.New.(*watchers.NetworkDefinition)
+				s.networkDefinitions[netDef.Name] = netDef
```
sknat, line 421: There is potential race-condition on `networkDefinitions` as it is accessed by two goroutines (this one and the CNI calls)
A way around this would be to `s.lock.Lock()` & `s.lock.Unlock()` around the accesses to it as it is also taken by `AddVppInterface()` and `DelVppInterface()`

sknat:
Hi @hedibouattour, thanks a lot for this PR ! I commented in a long overdue review, it's mostly nitpicks, and suggestions to make the logic simpler. The only issue I found was the missing lock around `networkDefinitions`

calico-vpp-agent/cni/cni_server.go
```diff
@@ -378,9 +405,32 @@ func (s *Server) ServeCNI(t *tomb.Tomb) error {
 	}
 
 	pb.RegisterCniDataplaneServer(s.grpcServer, s)
+
+	netsSynced := make(chan bool)
+	nets := make(map[string]*watchers.NetworkDefinition)
+	go func() {
+		for t.Alive() {
+			event := <-s.cniMultinetEventChan
```
sknat, line 413: We should probably also do a select
so that we don't wait on an empty chan on graceful teardown
````golang
select {
    case <-t.Dying():
        return
    case event := <-s.cniMultinetEventChan
        ...
} 
````

File: calico-vpp-agent/watchers/net_watcher.go (hunk 2 above)
Comment (hedibouattour) [reply calico-vpp-agent/watchers/net_watcher.go:90]: for is inside `w.watching(watcher)`

File: calico-vpp-agent/watchers/net_watcher.go (hunk 11 above)
Comment (hedibouattour) [reply calico-vpp-agent/watchers/net_watcher.go:83]: When I send netsSynced to the cni server it needs to make sure it knows all networks and sync its map. I think the way you are suggesting does not insure that all `NetAdded` events have arrived and all networks are synced in cni server at the moment `netsSynced` arrives.

Comment (hedibouattour) [reply calico-vpp-agent/watchers/net_watcher.go:83]: or are we sure given that it is the same channel so when netSynced arrives all previous messages have already arrived?

Comment (sknat) [reply calico-vpp-agent/watchers/net_watcher.go:83]: Yes, the channel guarantees the ordering, so yes I think we can rely on this to be sure that when `netSynced` arrives, all the previous messages have been handled.
The thing that worries me a bit with passing the map over the chan is that it's fairly easy to forget about it in the future & start messing with the same struct in both places

File: calico-vpp-agent/watchers/net_watcher.go (hunk 2 above)
Comment (sknat) [reply calico-vpp-agent/watchers/net_watcher.go:90]: Sorry I wasn't clear 😄 
I meant having a `for {}` loop instead of
````golang
restart:
some_logic()
goto restart
````

File: calico-vpp-agent/common/common.go [hunk 15]
```diff
@@ -201,13 +204,31 @@ func MakePath(prefix string, isWithdrawal bool, nodeIpv4 *net.IP, nodeIpv6 *net.
...
-	nlri, err := ptypes.MarshalAny(&bgpapi.IPAddressPrefix{
-		Prefix:    p.String(),
-		PrefixLen: uint32(masklen),
-	})
-	if err != nil {
-		return nil, err
+	var nlri *anypb.Any
+	if vni != 0 {
+		rdAttr, err := ptypes.MarshalAny(&bgpapi.RouteDistinguisherTwoOctetAS{
+			Admin:    64512,
```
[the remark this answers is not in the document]
hedibouattour: forgot it at hardcoded value, it should be the asNumber (passed as argument)

calico-vpp-agent/watchers/net_watcher.go
hedibouattour: I guess it is a low probability that this exceeds 64 char, we can keep it for now

sknat (approved):

PR #348, opened 2022-03-11
agent: add go tests

hedibouattour:

PR #390, opened 2022-06-01
Integration tests for CNI testing addition of nodes and pods

fgschwan:
This PR is based on https://github.com/projectcalico/vpp-dataplane/pull/348 . It adds integration test using running VPP instance and verifying CNI functionality correctly configuring it for event of adding new node in k8s (mentioned PR is adding test for adding new pod in k8s). The original PR was working with running k8s cluster, however the true dependency was only the VPP binary. The new PR is a rework(and test extension) using just docker container with VPP binary and there is no need to spin up the full k8s cluster.

PS: it seems that the master branch requires golang 1.17 instead of indicated golang 1.16 in go.mod file -> goland 1.17 was used in this PR


calico-vpp-agent/connectivity/ipsec.go
```diff
@@ -122,7 +122,7 @@ func (p *IpsecProvider) RescanState() {
 
 		p.log.Infof("Using async workers for ipsec, nbDataThread=%d", p.nDataThreads)
 		for i := 0; i < p.nDataThreads; i++ {
-			err = p.vpp.SetCryptoWorker(uint32(i), false)
+			err = p.vpp.SetCryptoWorker(uint32(i), false) // FIXME is this correct? It seems like disabling instead of enabling worker threads for IPSec
```
sknat, line 125: Agreed, this is quite unclear. The nDataThreads is set by `FetchNDataThreads()` in `common.go` but we probably should do a better job at naming it.

The idea when using IPSEC, is that we use the first nDataThread workers for RX/TXing packets, and the rest for async crypto. So we need to disable async crypto on the first "DataThreads"

test/integration-tests/run-tests.sh
```diff
@@ -0,0 +1,56 @@
...
+
+result=0
+
+echo "Running Integration tests..."
+echo "Running Calico VPP Agent - CNI tests..."
+prepageImageWithVPPBinary
+preparePodMockImage
+# Note: some pod tests expect elevated user privileges -> using sudo
+INTEGRATION_TEST="." VPP_IMAGE=$VPP_IMAGE VPP_BINARY="/usr/bin/vpp" \
+sudo -E env "PATH=$PATH" go test -v -run Integration ../../calico-vpp-agent/cni -ginkgo.v || result=$?
```
sknat, line 46: The ideal setup in which to run the tests is still a bit unclear to me. In the context of a CI, we'd probably want to run everything in a dedicated container (so without sudo) but for development it's probably worth having the option to sudo it...

calico-vpp-agent/connectivity/connectivity_server.go
```diff
@@ -352,3 +352,35 @@ func (s *ConnectivityServer) updateIPConnectivity(cn *common.NodeConnectivity, I
...
+func (s *ConnectivityServer) ForceProviderEnableDisable(providerType string, enable bool) (err error) {
+	provider, found := s.providers[providerType]
+	if !found {
+		return fmt.Errorf("can't find connectivity provider of type %s", providerType)
+	}
+	provider.EnableDisable(enable)
+	return nil
+}
+
+// TODO get rid (if possible) of all this "Force" methods by refactor the test code
```
sknat, line 379: I think it's ok for now to keep those around. It's not ideal, but I don't see an easy way to do introspection from the test package otherwise.

sknat (approved):
This looks really great ! I think we can go ahead and merge it, and build upon it to fill in the gas (deletion, updates, ...)

calico-vpp-agent/connectivity/ipsec.go
sknat: Thanks, yes I think this makes sense. It should be clearer in the context of ipsec

calico-vpp-agent/connectivity/connectivity_server.go
fgschwan: > I don't see an easy way to do introspection from the test package otherwise.

As i wrote on the second line of that TODO, it could be done by just running the event loop and sending event messages. Basically that way is the code running in the wild. The 3 "force" methods can be removed that way:
1. execute provider.RescanState() by simple run of the event loop function (it does the rescanstate for all providers)
2. execute s.providers[WIREGUARD].EnableDisable(...) by sending proper common.FelixConfChanged event
3. modify nodeByAddr by sending common.PeerNodeStateChanged event

Do we need to run event loop in the background? Will it complicate tests? Does it need additional work? Yes, yes and yes. But we can get rid of the ugly "force" method.

Basically it is a tradeoff between nicer functionality implementation and easier test implementation.
So no "easy" way, but it is doable.

fgschwan:
The test could be launched locally just by using the project-root makefile:
`make run-integration-tests`

sknat (approved):

PR #395, opened 2022-06-27
move vpp-manager watchers to agent and better support for multiuplink

hedibouattour:


calico-vpp-agent/vpp_manager_watchers/pool_sync.go
```diff
@@ -34,11 +34,13 @@ import (
 
 type PoolWatcher struct {
 	RouteWatcher *RouteWatcher
-	params       *config.VppManagerParams
-	conf         *config.LinuxInterfaceState
 
 	watcher              watch.Interface
 	currentWatchRevision string
+	UserSpecifiedMtu int
+	Mtu int
```
fgschwan, line 41: This line tells me that you are not using gofmt (or even better goimports) tool. It is quite common to format the go code with it as it is the code format standard that has support in standard go tooling. This repo doesn't have it as requirement for contributions so no need to apply it. However, for someone using it (i.e. in IDE), it could create unnecessary ghost changes (just formatting) in commits that can be disruptive (for committer if he/she wants to get rid of it before commit or for reviewer if committer doesn't care about formatting ghost changes)

vpp-manager/startup/interface_config.go
```diff
@@ -113,6 +113,10 @@ func loadInterfaceConfigFromLinux(ifSpec config.InterfaceSpec) (*config.LinuxInt
 	conf.NumTxQueues = link.Attrs().NumTxQueues
 	conf.NumRxQueues = link.Attrs().NumRxQueues
 	conf.Mtu = link.Attrs().MTU
+	if ifSpec.IsMain {
+		config.Info.MainUplinkMtu = conf.Mtu
+		//utils.WriteInfoFile()
```
fgschwan, line 118: Do have the multiple instances of "//utils.WriteInfoFile()" around the code some meaning hidden meaning or it just need to be cleaned?

vpp-manager/utils/utils.go
```diff
@@ -192,11 +198,13 @@ func CreateVppLink() (vpp *vpplink.VppLink, err error) {
...
 func ClearVppManagerFiles() error {
-	err := WriteFile("0", config.VppManagerStatusFile)
-	if err != nil {
-		return err
-	}
-	return WriteFile("-1", config.VppManagerTapIdxFile)
+	config.Info.Status = 0
+	config.Info.Uplinks = make(map[int]int)
+	config.Info.LinkMap = make(map[int]*config.Link)
+	//config.Info.ServiceCIDRs = make([]net.IPNet, 10)
```
fgschwan, line 204: Remove if not needed?

fgschwan (changes requested):

calico-vpp-agent/common/common.go
```diff
@@ -77,15 +79,19 @@ func CreateVppLink(socket string, log *logrus.Entry) (vpp *vpplink.VppLink, err
...
+func WaitForVppManager() (*vppmanagerconfig.VppManagerInfo, error) {
+	vppManagerInfo := &vppmanagerconfig.VppManagerInfo{}
 	for i := 0; i < 20; i++ {
-		dat, err := ioutil.ReadFile(config.VppManagerStatusFile)
-		if err == nil && strings.TrimSpace(string(dat[:])) == "1" {
-			return nil
+		dat, err := ioutil.ReadFile(config.VppManagerInfoFile)
+		if err == nil {
+			err2 := json.Unmarshal(dat, vppManagerInfo)
+			if err2 == nil && vppManagerInfo.Status == 1 {
```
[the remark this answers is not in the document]
hedibouattour: May be not fix the unmarshalling failure, but the fact that status is not equal to one, so both conditions need to be separated indeed, in order to return unmarshal failure fast, if any.

calico-vpp-agent/vpp_manager_watchers/pool_sync.go
hedibouattour: thanks, just forgot to run the formatter ;)

vpp-manager/startup/interface_config.go
hedibouattour: forgotten 😬

vpp-manager/config/config.go
```diff
@@ -51,6 +49,24 @@ const (
...
+
+type Link struct {
+	Name string
+	MTU  int
+}
+
+type VppManagerInfo struct {
+	Uplinks          map[int]int //map tapswifindex to linuxmtu
+	Status           int
+	UserSpecifiedMtu int
```
sknat, line 62: As `UserSpecifiedMtu` is the value of `getEnv("CALICOVPP_TAP_MTU")` we should probably pass it in the yaml instead of the json interface status

vpp-manager/config/config.go
```diff
@@ -51,6 +49,24 @@ const (
...
+var Info = &VppManagerInfo{}
+
+type Link struct {
+	Name string
+	MTU  int
+}
+
+type VppManagerInfo struct {
+	Uplinks          map[int]int //map tapswifindex to linuxmtu
+	Status           int
```
sknat, line 61: instead of an int `Status` could maybe be something more human-readable e.g. `ready` / `starting` the previous 0/1 was not super clear

vpp-manager/config/config.go
```diff
@@ -51,6 +49,24 @@ const (
...
+	UserSpecifiedMtu int
+	MainUplinkMtu    int //this mtu is used in routes for added networks
+	LinkMap          map[int]*Link
```
sknat, line 64: This should maybe be a list e.g. `UplinkStatuses: []UplinkStatus` with 
````go
type UplinkStatus struct {
    SwIfIndex uint32
    LinkIndex int
    Name int
    IsMain bool
    Mtu int
}
````

File: vpp-manager/config/config.go [hunk 8]
```diff
@@ -51,6 +49,24 @@ const (
...
 
+var Info = &VppManagerInfo{}
+
+type Link struct {
+	Name string
+	MTU  int
+}
+
+type VppManagerInfo struct {
+	Uplinks          map[int]int //map tapswifindex to linuxmtu
```
sknat, line 60: Same as below for `Uplinks` we could make this part of `UplinkStatuses`

vpp-manager/uplink/af_xdp.go
```diff
@@ -84,10 +84,14 @@ func (d *AFXDPDriver) PreconfigureLinux() error {
 			return errors.Wrapf(err, "Error reducing MTU to %d", maxAfXDPMTU)
 		}
 		d.conf.Mtu = maxAfXDPMTU
+		if d.spec.IsMain {
+			config.Info.MainUplinkMtu = maxAfXDPMTU
```
sknat, line 88: We should probably reduce the Mtu of `config.Info.UplinkStatus[X]`

calico-vpp-agent/config/config.go
```diff
@@ -139,12 +139,19 @@ func getEnvValue(str string) string {
...
+		dat, err := ioutil.ReadFile(VppManagerInfoFile)
 		if err == nil {
-			idx, err := strconv.ParseInt(strings.TrimSpace(string(dat[:])), 10, 32)
-			if err == nil && idx != -1 {
-				return int(idx), nil
+			err2 := json.Unmarshal(dat, vppManagerInfo)
+			if err2 == nil && len(vppManagerInfo.Uplinks) != 0 {
+				minMtu := 1000000
+				for _, v := range vppManagerInfo.Uplinks {
+					if v < minMtu {
```
sknat, line 150: But this logic should probably go in the `GetUplinkMtu()` function (json will default to 0, so we probably want to log an error and use `vpplink.MAX_MTU` if this happens)

calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -93,10 +96,47 @@ func main() {
 	}
 	// Once we have the api connection, we know vpp & vpp-manager are running and the
 	// state is accurately reported. Wait for vpp-manager to finish the config.
-	err = common.WaitForVppManager()
+	vppManagerInfo, err := common.WaitForVppManager()
 	if err != nil {
 		log.Fatalf("Vpp Manager not started: %v", err)
 	}
+	routeWatcher := &watchers.RouteWatcher{}
+	poolWatcher := &watchers.PoolWatcher{
```
sknat, line 104: This is probably worth a `NewPoolWatcher(...)` and `NewRouteWatcher()` to hide the logic.

calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -93,10 +96,47 @@ func main() {
...
+		RouteWatcher:     routeWatcher,
+		UserSpecifiedMtu: vppManagerInfo.UserSpecifiedMtu,
+		Mtu:              vppManagerInfo.MainUplinkMtu,
+		FakeNextHopIP4:   vppManagerInfo.FakeNextHopIP4,
+		FakeNextHopIP6:   vppManagerInfo.FakeNextHopIP6,
+	}
+	go routeWatcher.WatchRoutes()
```
sknat, line 111: We should probably stick to the `Go(routeWatcher.WatchRoutes)` syntax and have the tomb as a param to handle graceful teardown

calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -93,10 +96,47 @@ func main() {
...
+		err = routeWatcher.AddRoute(&netlink.Route{
+			Dst:      &serviceCIDR,
+			Gw:       gw,
+			Protocol: syscall.RTPROT_STATIC,
+			MTU:      watchers.GetUplinkMtu(vppManagerInfo.UserSpecifiedMtu, vppManagerInfo.MainUplinkMtu, true /* includeEncap */),
+		})
+		if err != nil {
+			log.Error(err, "cannot add tap route to service %s", serviceCIDR.String())
+		}
+	}
```
sknat, line 130: This could also go into the `linkWatcher.Serve()`

calico-vpp-agent/watchers/pool_sync.go
```diff
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package watchers
```
sknat, line 16: This could be done by listening to the events generated by the ipam watcher (e.g. `IpamConfChanged`) in fact it could even be done as part of the `uplink_route_watcher` and that way we could get rid of the whole locking logic

calico-vpp-agent/config/config.go
```diff
@@ -139,12 +139,19 @@ func getEnvValue(str string) string {
...
 func fetchHostMtu() (mtu int, err error) {
+	vppManagerInfo := &vppmanagerconfig.VppManagerInfo{}
 	for i := 0; i < 20; i++ {
-		dat, err := ioutil.ReadFile(VppManagerLinuxMtu)
+		dat, err := ioutil.ReadFile(VppManagerInfoFile)
 		if err == nil {
-			idx, err := strconv.ParseInt(strings.TrimSpace(string(dat[:])), 10, 32)
-			if err == nil && idx != -1 {
-				return int(idx), nil
+			err2 := json.Unmarshal(dat, vppManagerInfo)
```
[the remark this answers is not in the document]
hedibouattour: This actually is called before `waitForVppManager` so before setting the `common.VppManagerInfo` variable

vpp-manager/config/config.go
hedibouattour: you mean we add an env var `CALICOVPP_TAP_MTU` to agent ?

calico-vpp-agent/config/config.go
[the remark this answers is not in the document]
sknat: But if I understand correctly, this is only to set `config.HostMtu` which in turn is only used in `computePodMtu`
We should probably remove `config.HostMtu` and directly use the value in `vppManagerInfo` (also probably take the min of all MTUs to define the podMtu in case of multiple uplinks)

vpp-manager/utils/utils.go
```diff
@@ -76,6 +77,11 @@ func WriteFile(state string, path string) error {
 	return nil
 }
 
+func WriteInfoFile() error {
+	file, _ := json.MarshalIndent(config.Info, "", " ")
```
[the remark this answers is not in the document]
sknat: No ?
I meant something like 
````go
file, err := json.MarshalIndent(config.Info, "", " ")
if err != nil { ... }
````

File: calico-vpp-agent/watchers/uplink_route_watcher.go [hunk 17]
```diff
@@ -34,6 +43,22 @@ type RouteWatcher struct {
 	stop              bool
 	lock              sync.Mutex
```
sknat, line 44: `RouteWatcher.lock` can probably go, we seem to be now only using the resources it protects in a single loop

calico-vpp-agent/watchers/uplink_route_watcher.go
```diff
@@ -34,6 +43,22 @@ type RouteWatcher struct {
 	stop              bool
```
sknat, line 43: `RouteWatcher.stop` should probably be changed to leverage the `tomb` passed to `WatchRoutes()`

calico-vpp-agent/watchers/uplink_route_watcher.go
```diff
@@ -135,18 +160,75 @@ func (r *RouteWatcher) safeAddrClose() {
...
+		MTU:      GetUplinkMtu(config.UserSpecifiedMtu, true /* includeEncap */),
+	}, nil
+}
+
+func (r *RouteWatcher) WatchRoutes(t *tomb.Tomb) error {
 	r.netlinkFailed = make(chan struct{}, 1)
 	r.addrUpdate = make(chan struct{}, 10)
 	r.stop = false
 
 	go r.watchAddresses()
```
sknat, line 208: e.g. we could `r.watchAddresses(t)` instead of relying on `r.stop`

calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -93,12 +94,24 @@ func main() {
...
-	err = common.WaitForVppManager()
+	common.VppManagerInfo, err = common.WaitForVppManager()
 	if err != nil {
 		log.Fatalf("Vpp Manager not started: %v", err)
 	}
-
 	common.ThePubSub = common.NewPubSub(log.WithFields(logrus.Fields{"component": "pubsub"}))
+	routeWatcher := watchers.NewRouteWatcher()
+
+	Go(routeWatcher.WatchRoutes)
```
sknat, line 104: We could probably move those watchers next to the other ones (a.k.a. instantiation next to `NewNodeWatcher()` and startup next to `Go(nodeWatcher.WatchNodes)`)

calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -93,12 +94,24 @@ func main() {
...
+	var linkWatcher *watchers.LinkWatcher
+
+	if config.UserSpecifiedMtu != 0 {
+		linkWatcher = &watchers.LinkWatcher{
+			UplinkStatuses: common.VppManagerInfo.UplinkStatuses,
+		}
+	}
+	if linkWatcher != nil {
+		go linkWatcher.WatchLinks()
+	}
```
sknat, line 114: What about 
````go
linkWatcher := watchers.NewLinkWatcher(common.VppManagerInfo.UplinkStatuses)
if config.UserSpecifiedMtu != 0 {
     Go(linkWatcher.WatchLinks)
}
````
? (that way linkWatcher is never `nil`)

File: calico-vpp-agent/cmd/calico_vpp_dataplane.go [hunk 22]
```diff
@@ -227,7 +240,10 @@ func main() {
 	case <-t.Dying():
 		log.Errorf("tomb Dying %s", t.Err())
 	}
-
+	routeWatcher.Stop()
+	if linkWatcher != nil {
+		linkWatcher.Stop()
+	}
```
sknat, line 246: We could use the tomb to handle graceful teardown instead of calling these method

calico-vpp-agent/watchers/uplink_route_watcher.go
```diff
@@ -34,6 +43,22 @@ type RouteWatcher struct {
...
+	routeEventChan    chan common.CalicoVppEvent
+	FakeNextHopIP4    net.IP
+	FakeNextHopIP6    net.IP
+}
+
+func NewRouteWatcher() *RouteWatcher {
+	routeWatcher := &RouteWatcher{
+		routeEventChan: make(chan common.CalicoVppEvent, common.ChanSize),
+		FakeNextHopIP4: common.VppManagerInfo.FakeNextHopIP4,
+		FakeNextHopIP6: common.VppManagerInfo.FakeNextHopIP6,
```
sknat, line 55: Maybe we could pass it as a parameter of `NewRouteWatcher()` (same as for the LinkWatcher), just to make it clear that we depend on those values being set in the caller, and that it needs to happen after `WaitForVppManager()` is called

sknat (commented):
Thanks @hedibouattour for the changes,
just a few nitpicks, on some potential refactoring for the uplink watcher logics

vpp-manager/utils/utils.go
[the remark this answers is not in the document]
hedibouattour: Oh yes sorry I thought you meant the call to `WriteInfoFile`

sknat (approved):

PR #425, opened 2022-09-20
Add RX/TX placement support on memif interface

ondrej-fabry:
This PR introduces handling of interface events from VPP. It is used for setting RX/TX queues on workers for memif interface, because the interface must have link state up before this.


vpplink/vpplink.go
```diff
@@ -53,12 +58,22 @@ func NewVppLink(socket string, logger *logrus.Entry) (*VppLink, error) {
...
+	vppLink := &VppLink{
 		conn:   conn,
 		ch:     ch,
 		socket: socket,
 		log:    logger,
-	}, nil
+		pid:    uint32(os.Getpid()),
+	}
+
+	stopEvents, err := vppLink.watchInterfaceEvents()
```
sknat, line 69: Maybe we want to expose this function to the user (a.k.a require running `vpp.WatchInterfaceEvents()` if we want enabling it). This with the hypothesis that `vpplink/` is shared with other projects

ondrej-fabry: Yeah, was thinking that too.. for the purpose of avoiding enabling it by default. However it is only needs to be enabled once so perhaps I can make it start automatically on the first call to [`WatchInterfaceEvents`](https://github.com/ondrej-fabry/vpp-dataplane/blob/link-status/vpplink/interfaces.go#L771) and stop it automatically when last watcher finishes.

ondrej-fabry: That way user does not need to do extra call.. which might actually return error if the user calls it multiple times.

ondrej-fabry:
The update version now handles enable/disable procedures of event processing automatically. 
- when the **first watcher starts** - subscribes to interface events and enables them (sends `WantInterfaceEvents{Enable: 1}` message to VPP)
- and when the **last watcher stops** - disables interface events (sends `WantInterfaceEvents{Enable: 0}` message to VPP) and unsubscribes from interface events

ondrej-fabry:
CC @hedibouattour

sknat (approved):
lgtm, thanks @ondrej-fabry

PR #456, opened 2022-11-09
vpp-manager: Fix corefile cleanup

sknat:
This patch fixes a segfault happening when the corefile was not specified in the environment variable. It also tries to improve the logic a bit and adds a test for this function.

Additionnally this patch also enables running tests in `make test` and in the CI. This should not include the integration tests that require the envvar INTEGRATION_TEST to be set.

### DIFF
--- a/config/startup/startup.go
+++ b/config/startup/startup.go
@@ -67,7 +67,10 @@ const (
 )
 
 const (
-	/* Allow a maximum number of corefiles, delete older ones */
+/* Allow a maximum number of corefiles, delete older ones */
+)
+
+var (
 	maxCoreFiles = 2
 )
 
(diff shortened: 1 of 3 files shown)


config/startup/startup.go
```diff
@@ -304,36 +307,42 @@ func PrepareConfiguration(params *config.VppManagerParams) (conf []*config.Linux
...
+			return errors.Wrap(err, "Inner walk errored")
+		}
+		if !info.IsDir() {
+			timeAndPaths = append(timeAndPaths, timeAndPath{path, info.ModTime()})
 		}
 		return nil
 	})
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "walk errored")
```
hexotic, line 336: Minor: Wrapf without format...if the linter didn't raise a warning, don't bother changing it.

config/startup/startup_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	err = CleanupCoreFiles(filepath.Join(dir, "vppcore.%e.%p"))
+	if err != nil {
+		t.Errorf("Error calling CleanupCoreFiles %s", err)
+	}
+	for i := 0; i < 2; i++ {
+		_, err = os.Stat(filepath.Join(dir, fmt.Sprintf("file%d", i)))
+		if !os.IsNotExist(err) {
+			t.Errorf("file0 err is not ErrNotExist %s", err)
```
hexotic, line 51: file%d i ?

config/startup/startup_test.go
```diff
@@ -0,0 +1,71 @@
...
+package startup
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
```
hexotic, line 25: Maybe you could add a small comment to tell the reader what this function is about ;) ?

hexotic (approved):
Thanks for modifications

PR #459, opened 2022-11-22
Feature/bgp secrets

onong:
BGP password support as described in:

https://docs.tigera.io/master/security/comms/secure-bgp


calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -235,40 +256,63 @@ func (w *PeerWatcher) resyncAndCreateWatcher(state map[string]*bgpPeer) error {
...
-				err := w.deleteBGPPeer(ip)
+				err := w.deleteBGPPeer(ip, peer.BGPPeerSpec)
 				if err != nil {
 					return errors.Wrap(err, "error deleting BGP peer")
 				}
 				delete(state, ip)
 			}
 		}
+		// Clean up any secrets that are no longer referenced by any bgp peers
+		if w.secretWatcher != nil {
```
sknat, line 307: I'm not sure why we need this mark and sweep step. Couldn't we just remove the secret watcher when the parent BGPpeer is removed ? That probably requires a RefCnt in case multiple BGPPeers watch the same secret, but that'll

calico-vpp-agent/watchers/secret_watcher.go
```diff
@@ -0,0 +1,190 @@
...
+	// The channel that we should write to when we no longer want this watch
+	stopCh chan struct{}
+
+	// Secret value
+	secret *v1.Secret
+}
+
+type SecretWatcherClient interface {
+	// this function is invoked upon add|update|delete of a secret
+	OnSecretUpdate(secret string)
```
sknat, line 55: Why not `OnSecretUpdate(old, new *v1.Secret)` ? If we pass both previous / new values and the whole secret object, that'll allow us to directly use the name & secret value, and behave differently depending on whether it's an addition, update or deletion

calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -21,25 +21,34 @@ import (
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 type PeerWatcher struct {
 	log      *logrus.Entry
 	clientv3 calicov3cli.Interface
 
+	// Subcomponent for accessing and watching secrets (that hold BGP passwords).
+	secretWatcher          *secretWatcher
+	secretWatcherEventChan chan string
```
sknat, line 50: We can probably use the `peerWatcherEventChan` instead of creating a new chan for secrets

calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -396,12 +447,34 @@ func (w *PeerWatcher) SetBGPConf(bgpConf *calicov3.BGPConfigurationSpec) {
...
+		secretKeySelector.Name,
+		secretKeySelector.Key,
+	)
+	return password, err
+}
+
+// This function gets called from SecretWatcher when a secret is added, updated or deleted
+func (w *PeerWatcher) OnSecretUpdate(secret string) {
+	select {
+	case w.secretWatcherEventChan <- secret:
```
sknat, line 462: You can just write
````go
w.secretWatcherEventChan <- secret
````

Comment (sknat) [reply calico-vpp-agent/watchers/peers_watcher.go:462]: Also if we us the `peerWatcherEventChan` 
this can become 
````go
func OnSecretUpdate(old, new *v1.Secret) {
    common.SendEvent(common.CalicoVppEvent{
        Type: common.BGPSecretChanged,
        Old: old,
        New: new,
    })
}
````
this also needs to add the event type to the filter
````go
reg.ExpectEvents(common.PeerNodeStateChanged, common.BGPSecretChanged)
````

File: calico-vpp-agent/watchers/peers_watcher.go [hunk 5]
```diff
@@ -181,6 +192,15 @@ func (w *PeerWatcher) WatchBGPPeers(t *tomb.Tomb) error {
 			default:
 				goto restart
 			}
+		case secret := <-w.secretWatcherEventChan:
+			/* BGP secret related event - update the SecretUpdated field of affected peers */
+			w.log.Debugf("secret update event received for %s", secret)
+			for _, peer := range state {
+				if peer.BGPPeerSpec.Password != nil && peer.BGPPeerSpec.Password.SecretKeyRef != nil && peer.BGPPeerSpec.Password.SecretKeyRef.Name == secret {
+					w.log.Debugf("SecretUpdated field set for peer=%s", peer.BGPPeerSpec.PeerIP)
+					peer.SecretUpdated = true
```
sknat, line 201: Instead of punting the update of the password to the resync step, we can probably call `w.updateBGPPeer()` directly here for the peers that reference this Secret.

calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -235,40 +256,63 @@ func (w *PeerWatcher) resyncAndCreateWatcher(state map[string]*bgpPeer) error {
...
 				if ok {
+					w.log.Debugf("peer(update) neighbor ip=%s for BGPPeer=%s", ip, peer.ObjectMeta.Name)
 					existing.SweepFlag = false
-					if existing.AS != asn {
-						existing.AS = asn
-						err := w.updateBGPPeer(ip, asn)
+					oldSecret := ""
+					if existing.BGPPeerSpec.Password != nil && existing.BGPPeerSpec.Password.SecretKeyRef != nil {
+						oldSecret = existing.BGPPeerSpec.Password.SecretKeyRef.Name
+					}
```
sknat, line 264: We should probably have a function for this logic, as we're doing it quite a few times

calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -283,6 +303,16 @@ func (w *PeerWatcher) resyncAndCreateWatcher(state map[string]*bgpPeer) error {
 				delete(state, ip)
 			}
 		}
+		// Clean up any secrets that are no longer referenced by any bgp peers
+		if w.secretWatcher != nil {
```
sknat, line 307: We should always have `w.secretWatcher != nil` no ?

calico-vpp-agent/watchers/secret_watcher.go
onong: Yes, it would make sense if we want to do very elaborate processing with them. But our usage is rather simple really, namely, re-check/re-evaluate the BGP peers when there's some update related to the secret. For all the cases, ie, addition/deletion/update, there's really no difference in the processing required imo (at the moment, at least). Its the same in calico case too afaik. We could re-visit if need be, of course.

Moreover, the pair of (old, new) objects are available only for the update case, not for addition and deletion.

calico-vpp-agent/watchers/peers_watcher.go
onong: > You can just write
> 
> ```go
> w.secretWatcherEventChan <- secret
> ```

Thanks :)

calico-vpp-agent/watchers/peers_watcher.go
onong: Indeed :)

calico-vpp-agent/watchers/peers_watcher.go
onong: Agree

calico-vpp-agent/watchers/peers_watcher.go
```diff
@@ -178,6 +188,48 @@ func (w *PeerWatcher) WatchBGPPeers(t *tomb.Tomb) error {
...
+					switch secretEvt {
+					case "add":
+						// Note: any future add event specifc processing code goes here. For now we fallthrough.
+						fallthrough
+					case "del":
+						// Note: any future delete event specifc processing code goes here. For now we fallthrough.
+						fallthrough
+					case "upd":
+						// BGP password has changed
+						if w.getSecretName(peer.BGPPeerSpec) == secretName {
```
sknat, line 225: Couldn't we update the `Peer` object here ? (instead of setting `SecretChanged` to true )

sknat (commented):
lgtm, just a minor comment

calico-vpp-agent/watchers/peers_watcher.go
onong: I didn't quite understand. You mean the `peer.BGPPeerSpec.Password` field?

calico-vpp-agent/watchers/peers_watcher.go
sknat: Same, we can probably do this in a further refactoring

sknat (approved):

PR #463, opened 2022-11-28
yaml: add kind deployment without hugepages

hedibouattour:


yaml/overlays/dev/calico-vpp-daemonset-dev-patch-nohuge.yaml
```diff
@@ -0,0 +1,68 @@
...
+      volumes:
+        - name: repo-directory
+          hostPath:
+            path: ${REPO_DIRECTORY}
+        - name: ssl-certs
+          hostPath:
+            path: /etc/ssl/certs/
+        - name: share-certs
+          hostPath:
+            path: /usr/share/ca-certificates
```
sknat, line 68: If we are doing a patchmerge, we can probably leave out this parts if they do not change

sknat (approved):

PR #475, opened 2022-12-20
vpp-manager: Add uplink annotations, default disable qdisc bypass

sknat:
This adds an annotation map on the uplinkConfiguration object in the json definition allowing to change v2/v3 version for af_packet and the use of qdisc bypass or not.

We now default to qdisc bypass disables as this seems to be related to issues on EC2. But this is now exposed in the configuration.


onong (approved):
lgtm

PR #483, opened 2023-01-06
vpp-manager: fix corefile cleanup

sknat:
This patch fixes the coredump file cleanup only targeting the files that have the same prefix as the corepattern (to avoid deleting non coredump files)


hedibouattour (approved):

sknat:
Merged as part of #484

PR #487, opened 2023-01-12
Fix/policies issues

hedibouattour:


calico-vpp-agent/policy/rule.go
```diff
@@ -55,14 +55,14 @@ func (r *Rule) DeepCopy() *Rule {
...
-		SrcIPSetNames:          make([]string, len(r.DstIPPortSetNames)),
-		SrcNotIPSetNames:       make([]string, len(r.DstIPPortSetNames)),
+		DstIPPortIPSetNames:    make([]string, len(r.DstIPPortIPSetNames)),
+		DstNotIPPortIPSetNames: make([]string, len(r.DstNotIPPortIPSetNames)),
+		SrcIPPortIPSetNames:    make([]string, len(r.SrcIPPortIPSetNames)),
+		SrcNotIPPortIPSetNames: make([]string, len(r.SrcNotIPPortIPSetNames)),
+		DstIPSetNames:          make([]string, len(r.DstIPSetNames)),
+		DstNotIPSetNames:       make([]string, len(r.DstNotIPSetNames)),
+		SrcIPSetNames:          make([]string, len(r.SrcIPSetNames)),
+		SrcNotIPSetNames:       make([]string, len(r.SrcNotIPSetNames)),
```
sknat, line 65: 😮

hedibouattour: The bug was actually that policies with applyonforward=true were applied on tap and on uplink but with missing information on uplink, I found out that the issue was in the deepCopy of the rules but did not expect it to be this ^^

sknat (approved):

PR #496, opened 2023-01-18
agent: add egress policy allow from host to pods applied on tap

hedibouattour:


calico-vpp-agent/policy/policy_server.go
```diff
@@ -84,7 +84,9 @@ type Server struct {
 	workloadsToHostPolicy *Policy
 	workloadAddresses     map[string]string
 	/* always allow traffic coming from host to the pods (for healthchecks and so on) */
-	allowFromHostPolicyId uint32
+	IngressAllowFromHostPolicyId uint32
+	EgressAllowFromHostPolicyId  uint32
```
sknat, line 88: We can probably create a single `AllowFromHostPolicyId` containing both rules (RX & TX),
and assign it in Egress on the hep, and in Ingress on weps

calico-vpp-agent/policy/policy_server.go
```diff
@@ -84,7 +84,9 @@ type Server struct {
 	workloadsToHostPolicy *Policy
 	workloadAddresses     map[string]string
 	/* always allow traffic coming from host to the pods (for healthchecks and so on) */
```
sknat, line 86: We should probably be more descriptive in the comment & the function creating it.
e.g.
````golang
// allowFromHostPolicyId persist the ID of the policy allowing host <-> pod communications. 
// See CreateAllowFromHostPolicyId definition
allowFromHostPolicyId uint32
````
and 
````golang
// CreateEgressAllowFromHostPolicy creates a policy allowing host->pod communications. This is needed
// to maintain vanilla Calico's behavior where the host can always reach pods on any pods.
// this policy is applied in Egress on the host endpoint tap (i.e. linux -> VPP)
// and on the Ingress of Workload endpoints (i.e. VPP -> pod)
func CreateEgressAllowFromHostPolicy() { ... }
````

File: calico-vpp-agent/policy/policy_server.go [hunk 3]
```diff
@@ -84,7 +84,9 @@ type Server struct {
...
-	allowFromHostPolicyId uint32
+	IngressAllowFromHostPolicyId uint32
+	EgressAllowFromHostPolicyId  uint32
+	AllowFromHostIPSet           *IPSet
```
sknat, line 89: Same thing here, we should probably add a comment stating that this is the ipset for the AllowFromHostPolicy
````golang 
// AllowFromHostIPSet
AllowFromHostIPSet *IPSet
````

sknat (commented):
Looks good, just a potential simplification & comments that could be added 😄

sknat (approved):

sknat (approved):

PR #498, opened 2023-01-19
Fix manifest yamls

onong:
Fix the yamls which got messed up post the config changes.

1. Add back the missing fields for eks
2. Set vppDriver explicitly per individual manifests


sknat (approved):
Oh, good catch, thanks @onong !

PR #504, opened 2023-01-20
agent: add startup watchdog logging missing components

hedibouattour:
adding a startup watchdog that logs every 5 seconds what we are missing for agent to start (felixconfig, bgpspec)

### DIFF
--- a/calico-vpp-agent/cmd/calico_vpp_dataplane.go
+++ b/calico-vpp-agent/cmd/calico_vpp_dataplane.go
@@ -151,22 +151,18 @@ func main() {
 	routingServer.SetBGPConf(bgpConf)
 	serviceServer.SetBGPConf(bgpConf)
 
+	watchDog := NewWatchDog(log.WithFields(logrus.Fields{"component": "watchDog"}))
+	Go(watchDog.watch)
 	Go(policyServer.ServePolicy)
-	log.Infof("Waiting for our node's BGP spec...")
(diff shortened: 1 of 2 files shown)


calico-vpp-agent/cmd/calico_vpp_dataplane.go
```diff
@@ -151,23 +151,19 @@ func main() {
 	routingServer.SetBGPConf(bgpConf)
 	serviceServer.SetBGPConf(bgpConf)
 
+	watchDog := NewWatchDog(log.WithFields(logrus.Fields{"component": "watchDog"}))
+	Go(watchDog.watch)
 	Go(policyServer.ServePolicy)
-	log.Infof("Waiting for our node's BGP spec...")
 	felixConfig := policyServer.WaitForFelixConfig()
+	felixConfigReceived = true
```
sknat, line 158: Instead of relying on variables, we could do something like 
````go
watchDog.Wait(felixConfig, "Waiting for FelixConfig to be provided by the calico pod")
// with 
type WatchDogWatchable interface (
   func GetWaitChan() <-struct{}
)
````
That will allow disconnections between the variables and the `WaitForFelixConfig()` calls if refactoring happen at some point. Also we can do a select on this chan but also on the tomb (to return in case something fails)

hedibouattour: addressed in #505

sknat:
Closed in favor of #505

PR #505, opened 2023-01-20
refactor multinet documentation

hedibouattour:


calico-vpp-agent/policy/policy_server.go
```diff
@@ -1257,7 +1253,7 @@ func (s *Server) onNodeAdded(node *common.LocalNodeSpec) (err error) {
 		(node.IPv4Address != nil || node.IPv6Address != nil) {
 		if s.ip4 == nil && s.ip6 == nil {
 			/* We found a BGP Spec that seems valid enough */
-			s.gotOurNodeBGPchan <- *node
+			s.GotOurNodeBGPchan <- node
```
sknat, line 1256: This looks like it should stay `*node` no ?

hedibouattour: casting wouldn't work like that, `node` is already a pointer

sknat: It's not casting I'm afraid of, rather the fact that we pass a shared reference over to multiple readers, but I guess the previous implementation was not doing a great job either. 
We can probably rely on the fact that this should be a const from the point of view of the goroutines

sknat (approved):

PR #518, opened 2023-01-25
agent: fix endpointtohostaction policy when set to accept

hedibouattour:
new behavior is: we create a common ipset (as we use it in both endpointtohost and allowfromhost). When endpointtohostaction changes we create a rule with allow/deny targeting that ipset

### DIFF
--- a/calico-vpp-agent/policy/policy_server.go
+++ b/calico-vpp-agent/policy/policy_server.go
@@ -80,17 +80,14 @@ type Server struct {
 	/* failSafe policies allow traffic on some ports irrespective of the policy */
 	failSafePolicy *Policy
 	/* workloadToHost may drop traffic that goes from the pods to the host */
-	workloadsToHostIPSet   *IPSet
 	workloadsToHostPolicy  *Policy
 	defaultTap0IngressConf []uint32
-	workloadAddresses      map[string]string
 	/* always allow traffic coming from host to the pods (for healthchecks and so on) */
 	// AllowFromHostPolicy persists the policy allowing host --> pod communications.
 	// See CreateAllowFromHostPolicy definition
 	AllowFromHostPolicy *Policy
-	// AllowFromHostIPSet persists the ipset containing all the workload endpoints (pods)
-	// addresses, it is used to allow traffic from host to pods
-	AllowFromHostIPSet *IPSet
+	// allPodsIpset persists the ipset containing all the workload endpoints (pods) addresses
+	allPodsIpset *IPSet
 	/* allow traffic between uplink/tunnels and tap interfaces */
 	allowToHostPolicy *Policy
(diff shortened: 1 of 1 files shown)


calico-vpp-agent/policy/policy_server.go
```diff
@@ -1677,50 +1674,49 @@ func (s *Server) createAllowFromHostPolicy() (err error) {
...
 	}
-	ipset := NewIPSet()
-	ps := PolicyState{IPSets: map[string]*IPSet{"ipset1": ipset}}
-	err = ipset.Create(s.vpp)
-	if err != nil {
-		return err
-	}
-	pol.InboundRules = append(pol.InboundRules, r_deny_workloads)
-	err = pol.Create(s.vpp, &ps)
+	ps := PolicyState{IPSets: map[string]*IPSet{"ipset1": s.allPodsIpset}}
```
sknat, line 1692: `ipset1` ? Is this needed ? Just wondering whether this will conflict with another name, we could also leave it blank no ?

calico-vpp-agent/policy/policy_server.go
```diff
@@ -1677,50 +1674,49 @@ func (s *Server) createAllowFromHostPolicy() (err error) {
...
 
 func (s *Server) createEndpointToHostPolicy( /*may be return*/ ) (err error) {
-	pol := &Policy{
+	workloadsToHostPolicy := &Policy{
 		Policy: &types.Policy{},
 		VppID:  types.InvalidID,
 	}
-	r_deny_workloads := &Rule{
+	action := types.ActionDeny
+	if s.getEndpointToHostAction() == "ACCEPT" {
```
sknat, line 1682: `s.getEndpointToHostAction()` could directly return `types.ActionAllow` or `types.ActionDeny` that would allow to simplify part of the logic

sknat (commented):
A few comments, other than that lgtm

calico-vpp-agent/policy/policy_server.go
hedibouattour: it is just an identifier that should be the same as in the rule list of ipsets
The rule has the ipset named "ipset1", so it looks for this name in the policyState

sknat: Agreed, but this name comes from Calico, if at some point Calico generates `ipset1` as an ipset name, both will conflict

sknat: Also we could use something more descriptive than `ipset1` e.g. `calico-vpp-XXX-ipset` with `XXX` something like `wep-addr` ?

hedibouattour: fair

sknat (approved):

PR #536, opened 2023-02-09
vpp: foward mcast traffic to calico-pods-46

sknat:
This patch forwards the multicast traffic originiating from the pods to the calico-pods-46 VRF, so that this traffic might be configured by users without knowleged of the pod's VRF table_id.


AloysAugustin (approved):
lgtm

PR #554, opened 2023-02-28
multinet: fix loopbacks leftover state in vxlan provider

hedibouattour:
A bug appears when you: add a network, add pods in it (vxlan triggered), then delete the pods and network, then re-add it.
There is a leftover state of a configured loopback in vxlan provider that causes this.
Solution is to use the map that is updated when the network is created/deleted.

### DIFF
--- a/calico-vpp-agent/connectivity/vxlan.go
+++ b/calico-vpp-agent/connectivity/vxlan.go
@@ -33,12 +33,10 @@ type VXLanProvider struct {
 	vxlanRoutes  map[uint32]map[string]bool
 	ip4NodeIndex uint32
 	ip6NodeIndex uint32
-
-	netsLoopbacks map[uint32]bool
 }
 
 func NewVXLanProvider(d *ConnectivityProviderData) *VXLanProvider {
-	return &VXLanProvider{d, make(map[string]types.VXLanTunnel), make(map[uint32]map[string]bool), 0, 0, make(map[uint32]bool)}
+	return &VXLanProvider{d, make(map[string]types.VXLanTunnel), make(map[uint32]map[string]bool), 0, 0}
 }
 
 func (p *VXLanProvider) EnableDisable(isEnable bool) {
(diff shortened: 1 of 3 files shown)


calico-vpp-agent/connectivity/vxlan.go
```diff
@@ -141,8 +139,7 @@ func (p *VXLanProvider) AddConnectivity(cn *common.NodeConnectivity) error {
 		return err
 	}
 	if cn.Vni != 0 {
-		_, found := p.netsLoopbacks[cn.Vni]
-		if !found {
+		if !p.server.networks[cn.Vni].LoopbackConfigured {
```
sknat, line 142: Couldn't we move all this code to `calico-vpp-agent/cni/pod_interface/loopback.go:CreateInterface()` ? That would remove the need to add `LoopbackConfigured` on the network object and updating state here

hedibouattour: that one is specific for pod interface loopback, here we have a loopback for the secondary network tunnel. Plus `/cni/pod_interface` is specific for cni and pods. Do you mean we create a common function that creates loopbacks for pods interfaces and for networks vxlans and put somewhere in a common place?

sknat: Oh sorry, I meant in `CreateVRFsForNet()`, I got confused by searching for `LoopbackSwIfIndex =`

calico-vpp-agent/watchers/net_watcher.go
```diff
@@ -272,6 +280,19 @@ func (w *NetWatcher) CreateVRFsForNet(networkName string, networkVni uint32, net
...
 	w.networkDefinitions[networkName] = netDef
+
+	ip4, ip6 := w.GetNodeIPs()
+	var ipaddrs = []*net.IP{ip4, ip6}
+	for idx, ipFamily := range vpplink.IpFamilies {
+		err = w.vpp.SetInterfaceVRF(swIfIndex, netDef.VRF.Tables[idx], ipFamily.IsIp6)
+		if err != nil {
+			return nil, errors.Wrapf(err, "Error setting loopback %d in network vrf", swIfIndex)
+		}
+		err = w.vpp.AddInterfaceAddress(swIfIndex, common.ToMaxLenCIDR(*ipaddrs[idx]))
```
sknat, line 291: This will segfault if we don't have both `ip4` & `ip6` configured.

sknat (approved):

PR #559, opened 2023-03-01
vpp: memif, fix null termination of socket name

sknat:
The socket name was not properly null terminated when passed via the API. This should go away with the next rebase of VPP.


hexotic (approved):
LGTM

PR #563, opened 2023-03-03
build: pass TAG to integration tests & default to WITH_GDB=yes

sknat:
This default to WITH_GDB=yes to have GDB present in all the build as it is low footprint (~40Mb), and quite handy to troubleshoot VPP related issues.

This also makes the integration-tests script inherit the TAG from the makefile .ci/common.mk


hexotic (approved):
LGTM. We'll refactor later

PR #574, opened 2023-03-21
add scale tests

hedibouattour:


test/kind/new_cluster.sh
```diff
@@ -18,31 +18,63 @@ containerdConfigPatches:
...
+  k=$((i+2))
+fi
+
 config=$(cat <<EOF
 $config
 - role: worker
   extraMounts:
     - hostPath: $HOME
       containerPath: $HOME
+  cpuSet: "$j,$k"
```
sknat, line 77: Maybe we could have this directly in the script, e.g.
````bash
FIRST_CPU=3
N_KIND_WORKERS=4
for ((i=0; i<$N_KIND_WORKERS; i++)); do
cat <<EOF
cpuSet: "$((i+$FIRST_CPU)),$((i+$FIRST_CPU+2*((i+1)%2)-1))"
EOF
done
````

File: test/kind/Makefile [hunk 2]
```diff
@@ -18,24 +20,43 @@ dev:
 	docker image tag calicovpp/multinet-monitor:$(TAG) localhost:5000/calicovpp/multinet-monitor:$(TAG)
 	docker push localhost:5000/calicovpp/multinet-monitor:$(TAG)
 
+# this is for a cluster using a mirror docker.io -> localhost:5000 to avoid pull rate limit
+.PHONY: load-mirror
+load-mirror:
+	docker pull docker.io/calico/node:$(KIND_CALICO_VERSION)
```
sknat, line 26: Shouldn't we be using a list of image names (as we are using them in a few place, just to make sure we don't miss one in one of the directives). We're using a similar pattern for e.g. `make dev.k3s` in the top level Makefile

sknat (commented):
Hi @hedibouattour,
thanks for the PR, a few comments inline 😄

sknat (approved):
lgtm, 
but the CI seems unhappy 😕

sknat (approved):

PR #578, opened 2023-04-04
refact: move agent binaries to bin/

sknat:
This patch moves agent binaries to bin/ instead of the same folder as their sources.

### DIFF
--- a/calico-vpp-agent/Makefile
+++ b/calico-vpp-agent/Makefile
@@ -11,20 +11,23 @@ all: build gobgp image
 
 export GOOS=linux
 
+bin:
+	mkdir -p bin
+
 # We make felix-api-proxy a static executable as it will run in the calico container
 # for which we have less control on the env and glibc version
 .PHONY: felix-api-proxy
 felix-api-proxy: CGO_ENABLED=0
-felix-api-proxy:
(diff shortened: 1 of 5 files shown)


calico-vpp-agent/Makefile
```diff
@@ -11,20 +11,23 @@ all: build gobgp image
...
 felix-api-proxy: CGO_ENABLED=0
-felix-api-proxy:
-	${DOCKER_RUN} go build -o ./cmd/felix-api-proxy ./cmd/api-proxy
+felix-api-proxy: bin
+	${DOCKER_RUN} go build -o ./bin/felix-api-proxy ./cmd/api-proxy
 
-build: felix-api-proxy
-	${DOCKER_RUN} go build -o ./cmd/calico-vpp-agent ./cmd
-	${DOCKER_RUN} go build -o ./cmd/debug ./cmd/debug-state
+build: felix-api-proxy bin
```
hedibouattour, line 24: `felix-api-proxy` target already calls `bin`, I think this "bin" can be deleted

hedibouattour (approved):

calico-vpp-agent/Makefile
sknat: Agreed, just left it there in case at some point we decide to remove `felix-api-proxy`

PR #597, opened 2023-05-17
fix for DNS issue in OpenShift

onong:
NetworkManager wipes out DNS config when it sees that the interface has vanished/changed. The fix appends "dns=none" in the NetworkManager.conf file which prevents this calamity.


vpp-manager/startup/startup.go
```diff
@@ -115,13 +115,21 @@ func parseEnvVariables(params *config.VppManagerParams) (err error) {
 		hooks.RegisterBashHook(hooks.BEFORE_IF_READ, conf)
 	}
 
+	// NetworkManager wipes out DNS config when the uplink interface
+	// is moved. We counter it by telling it to leave the DNS config alone.
+	hooks.RegisterBashHook(hooks.BEFORE_VPP_RUN, hooks.FIX_DNS_SCRIPT)
```
sknat, line 120: Maybe that'd be easier with renaming `network_restart.sh` : `default_hook.sh` and passing the hookpoint when calling it.
That way we can start the script with
````bash
if [ $1 = "BEFORE_VPP_RUN" ]; then 
   ...
elif ....
````
and have everything in a single glance.

File: vpp-manager/startup/startup.go [hunk 2]
```diff
@@ -115,13 +115,21 @@ func parseEnvVariables(params *config.VppManagerParams) (err error) {
...
+
 	// Add default hook if none specified
 	for _, hookName := range []string{hooks.VPP_RUNNING, hooks.VPP_DONE_OK, hooks.VPP_ERRORED} {
 		if hooks.HookCount(hookName) == 0 {
 			hooks.RegisterBashHook(hookName, hooks.DEFAULT_RESTART_SCRIPT)
```
sknat, line 125: Having a single script would also allow user override with this function (we should probably add `BEFORE_VPP_RUN` to the lot of hooks having our script as the default)

sknat (commented):
overall lgtm,
Just a few comments regarding how defaults behave / where the logic goes

vpp-manager/hooks/default_hook.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/sh
+
+HOOK="$0"
```
sknat, line 3: Shouldn't it be `$1` ? (iirc `$0` is the script name)

vpp-manager/startup/startup.go
onong: Great idea. Thanks.

vpp-manager/hooks/default_hook.sh
sknat: Right, my bad 😕 
I didn't recall we were passing it inline

sknat (approved):

PR #614, opened 2023-06-26
yamls: transition to components

sknat:
This patch transitions the recursive overalys to using
kubectl kustomize components instead. This makes importing the
same yaml patch (e.g. srv6 ; multinet) in different manifests easier.

We also make sure that all the overlays defined build in the make yaml
directive, even if we do not version the test-* overlays.


yaml/components/ipsec/README.md
hedibouattour, line 1: I think we should have this in `docs/` as well, next to other documentations

yaml/generated/calico-vpp-multinet.yaml
hedibouattour, line 1: those were hardcoded because we don't have prerelease version, we should either add it as a tagged image or make it hardcoded

hedibouattour (commented):
For kind we have both `yaml/overlays/kind` and `yaml/overlays/dev`, should we address standardizing things in another PR?

yaml/generated/calico-vpp-multinet.yaml
sknat: > For kind we have both yaml/overlays/kind and yaml/overlays/dev, should we address standardizing things in another PR?

I think it's fair to have both : 
- We can have `/kind` be the generic kind installation (i.e. an external user wanting to install the latest release on a vanilla kind cluster), 
- On the other hand `/dev` is a really complex overlay which intends to support all the knobs we expose in Calico/VPP. It's true it leverages kind under the hood, but I think usecases are arguably different enough to keep both (or make dev be just an overlay on top of `/kind`

yaml/Makefile
```diff
@@ -4,13 +4,23 @@ CALICO_TAG ?= master
...
-	@kubectl kustomize overlays/eks-dpdk       > generated/calico-vpp-eks-dpdk.yaml
-	@kubectl kustomize overlays/kind           > generated/calico-vpp-kind.yaml
+	kubectl kustomize base                                     > generated/calico-vpp-nohuge.yaml
+	kubectl kustomize overlays/default-huge                    > generated/calico-vpp.yaml
+	kubectl kustomize overlays/default-multinet                > generated/calico-vpp-multinet.yaml
+	kubectl kustomize overlays/dpdk                            > generated/calico-vpp-dpdk.yaml
+	kubectl kustomize overlays/eks                             > generated/calico-vpp-eks.yaml
+	kubectl kustomize overlays/eks-dpdk                        > generated/calico-vpp-eks-dpdk.yaml
+	kubectl kustomize overlays/kind                            > generated/calico-vpp-kind.yaml
+	kubectl kustomize overlays/kind                            > generated/calico-vpp-kind.yaml
```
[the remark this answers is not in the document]
sknat: 😮

hedibouattour (approved):

PR #617, opened 2023-07-05
doc: add vcl doc

hedibouattour:


docs/vcl.md
```diff
@@ -0,0 +1,80 @@
+## Introduction and Overview
+
+VCL is a communication library in VPP that provides a socket-based interface for client/server communication with VPP. It is designed to enable client applications to establish connections with VPP, send and receive control messages, and perform configuration and management operations.
```
sknat, line 3: Maybe something like :

VCL stands for the [VPP Comms Library](https://wiki.fd.io/view/VPP/HostStack/VCL), it is a library that allows application to easily leverage the VPP hoststack (i.e. TCP, UDP, TLS, QUIC, ... stack implementation in VPP itself) in a similar way as the libc works for kernel hoststack syscalls. 
It works by exposing a unix socket over which control messages exchanged and configurations negotiated.

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: samplepod
+  annotations:
+    "cni.projectcalico.org/vppVcl": "enable"
+```
+This will trigger vpp to create the VCL socket and add a session, which is an abstraction that allows applications to send and receive data over the network.
```
sknat, line 37: This will only ask VPP to create the abstract socket, the session creation happens when you issue listens or connects calls over the VCL.
When we start the iperf container that's done by the iperf3 entrypoint that issues a listen over LDP+VCL

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Sockets
+
+VCL in VPP uses a socketfile: a Unix domain socket used for communication between endpoints. This allows sessions to initiate connections.
```
sknat, line 41: The VPP Hoststack exposes an abstract socket, it's a unix socket over which applications communicate with VPP - leveraging the VCL - in order to open or accept L4+ connections (e.g. TCP) directly in VPP.

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+We use an abstract socket for our interface. This socket address is usually represented as an abstract name beginning with a null byte followed by the desired name, typically "@mysocket".
```
sknat, line 42: Calico/VPP uses abstract socket, which being anchored to a network namespace are better suited for the CNI world rather than regular socket which are mounted on the filesystem. Their address are encoded as a name with a heading null byte, which is usually replaced by the `@` sign when printed.

Calico/VPP exposes the `@vpp/session` abstract socket within the container, to which the VCL should be pointed.

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+VPP api has a particular syntax for abstract sockets: using the keyword `abstract:` then `netns_name:`. 
+For example: `abstract:vpp/session,netns_name=/var/run/netns/cni-75e26661-4119-90a4-b9d2-0b00489f76f3`. This syntax appears in vppctl.
+
+## Troubleshooting vcl interface creation
+
+To check abstract socket creation in linux, you can run the following command on the pod running vcl:
+
+```bash
+lsof -U | grep session
```
sknat, line 51: We should probably also add the `ss` and `netstat` based equivalents

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+vpp_main 3046023 root   33u  unix 0xffff9eaeb1943740      0t0 396363922 @vpp/session type=SEQPACKET
+```
+
+To check sessions in vpp cli:
```
sknat, line 55: We should probably add a snippet on what sessions are (i.e. listeners or established connections in various protocols). Probably doing a parallel with the output of `ss` and `netstat` would help, before describing `show session verbose`
A warning should probably be added stating that the output might be huge on live systems with a lot of connections

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Testing VCL feature
+
+VCL can be tested using iperf network performance tool.
+here is a [deployment](../test/yaml/iperf3-vcl/test.yaml) example with client/server pods for tests.
+```bash
+test/scripts/vppdev.sh tst up iperf3-vcl
+test/scripts/vppdev.sh cases ipv4_vcl
+
+```
```
sknat, line 80: We should probably directly embed the yamls (as we already have a part of it above)
and add a section about what the entrypoint does in the iperf3-vcl case (the `iperf3-vcl.sh` script),
i.e. :
- generate the `/etc/vpp/vcl.conf` configuration file which will be picked up by the VCL library as we did set `VCL_CONFIG=/etc/vpp/vcl.conf`
- use `LD_PRELOAD=/usr/local/lib/vpp/libvcl_ldpreload.so` which is roughly equivalent of dynamically linking the iperf3 binary against the VCL instead of the libc (so piping the connect() and accept() syscall to VPP instead of linux)
- and then doing `iperf3 -4 -s` which creates the listener session
````
[0:1][T] 11.0.0.139:5201->0.0.0.0:0 LISTEN
````

Comment (sknat) [reply docs/vcl.md:80]: This session is probably also better put before the Troubleshooting one

File: docs/vcl.md [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+This will trigger vpp to create the VCL socket and add a session, which is an abstraction that allows applications to send and receive data over the network.
+
+### Sockets
+
+VCL in VPP uses a socketfile: a Unix domain socket used for communication between endpoints. This allows sessions to initiate connections.
+We use an abstract socket for our interface. This socket address is usually represented as an abstract name beginning with a null byte followed by the desired name, typically "@mysocket".
+VPP api has a particular syntax for abstract sockets: using the keyword `abstract:` then `netns_name:`. 
+For example: `abstract:vpp/session,netns_name=/var/run/netns/cni-75e26661-4119-90a4-b9d2-0b00489f76f3`. This syntax appears in vppctl.
+
+## Troubleshooting vcl interface creation
```
sknat, line 46: Maybe just `Troubleshooting VCL` ?

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: samplepod
+  annotations:
+    "cni.projectcalico.org/vppVcl": "enable"
+```
+This will trigger vpp to create the VCL socket and add a session, which is an abstraction that allows applications to send and receive data over the network.
+
+### Sockets
```
sknat, line 39: We can probably merge this with the previous section, as this falls into the "using VCL in pods" story no ?

docs/vcl.md
```diff
@@ -0,0 +1,80 @@
+## Introduction and Overview
```
sknat, line 1: We can probably add one section at the bottom :

### known issues
- VCL currently only works for the primary interface when used in combination with multinet
- Dataplane restart is currently not detected by the client application, this can be alleviated by making healthchecks travel over the VCL.
- Source address spoofing is not supported even when disabling uRPF at the pod level.

sknat (commented):
Thanks @hedibouattour for this docs section !
A few comments inline, also we could probably link to the Hoststack / VCL docs themselves at 
- https://wiki.fd.io/view/VPP/HostStack/VCL
- https://wiki.fd.io/view/VPP/HostStack

PR #622, opened 2023-07-13
agent: fix nodeport service snat

hedibouattour:
We snat when endpoint is not local, no need to do it when it is local.
This is for the returned traffic to pass by the node.


hedibouattour:
snat should be applied only in non-local endpoint cases, for nodeport services

sknat (approved):

PR #623, opened 2023-07-18
docs: Add kind doc for users

sknat:
This patch adds documentation on how to set up a kind cluster with calico/vpp & multinet installed, without requiring to build VPP locally or to have the whole repository cloned locally.

This should work without requiring to run any remote script.

### DIFF
--- a/docs/kind.md
+++ b/docs/kind.md
@@ -1,0 +1,74 @@
+
+# Setting up a test kind cluster 
+
+## Default installation
+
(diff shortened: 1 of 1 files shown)


hedibouattour:
Tested it, it works fine, lgtm.

sknat:
Probably worth adding a how to try kind without multinet

hedibouattour (approved):

PR #626, opened 2023-08-11
simplify punt fake gateway

sknat:
This patch simplifies the way we handle the punt routes when routing for-me traffic VPP cannot handle to the host. Previously, we were using different fake gateways for ip4 & ip6. We now use the ip4 indirection in both cases.
We also generate different hardware addresses using the uplink order in the configuration.

Finally, we also make the isMain parameter public in the config.

### DIFF
--- a/calico-vpp-agent/cni/cni_server.go
+++ b/calico-vpp-agent/cni/cni_server.go
@@ -38,7 +38,6 @@ import (
 	"github.com/projectcalico/vpp-dataplane/v3/calico-vpp-agent/common"
 	"github.com/projectcalico/vpp-dataplane/v3/calico-vpp-agent/watchers"
 	"github.com/projectcalico/vpp-dataplane/v3/config"
-	"github.com/projectcalico/vpp-dataplane/v3/vpp-manager/utils"
 	"github.com/projectcalico/vpp-dataplane/v3/vpplink"
 	"github.com/projectcalico/vpp-dataplane/v3/vpplink/types"
 )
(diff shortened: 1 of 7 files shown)


calico-vpp-agent/cni/cni_server.go
```diff
@@ -506,12 +503,15 @@ func (s *Server) createRedirectToHostRules() (uint32, error) {
...
+	mainInterface := s.getMainInterface()
+	if mainInterface == nil {
+		return types.InvalidID, fmt.Errorf("No main interface found")
+	}
 	for _, rule := range config.GetCalicoVppInitialConfig().RedirectToHostRules {
 		err = s.vpp.AddSessionRedirect(&types.SessionRedirect{
 			FiveTuple:  types.NewDst3Tuple(rule.Proto, net.ParseIP(rule.Ip), rule.Port),
 			TableIndex: index,
-		}, &types.RoutePath{Gw: tap0nexthop, SwIfIndex: tap0swifindex})
+		}, &types.RoutePath{Gw: mainInterface.FakeNextHopIP4, SwIfIndex: mainInterface.TapSwIfIndex})
```
hedibouattour, line 514: I think here we should rather use `config.VppHostPuntFakeGatewayAddress` for session redirection

hedibouattour (changes requested):

calico-vpp-agent/cni/cni_server.go
sknat: Oh, good catch, thanks !

PR #635, opened 2023-09-11
vpp: pass hashconfig to vpp for service annotations

hedibouattour:
This patch uses the newly introduced field in cnat translation vpp api to pass ecmp hashconfig defined by annotations in services.

### DIFF
--- a/calico-vpp-agent/services/service.go
+++ b/calico-vpp-agent/services/service.go
@@ -15,24 +15,18 @@
 
 package services
 
+import "github.com/projectcalico/vpp-dataplane/v3/vpplink/types"
+
 type lbType string
-type hashConfig string
 
 const (
-	lbTypeECMP          lbType     = "ECMP"
-	lbTypeMaglev        lbType     = "Maglev"
-	lbTypeMaglevDSR     lbType     = "MaglevDSR"
-	hashConfigSrcport   hashConfig = "srcport"
-	hashConfigDstport   hashConfig = "dstport"
-	hashConfigSrcaddr   hashConfig = "srcaddr"
-	hashConfigDstaddr   hashConfig = "dstaddr"
-	hashConfigIproto    hashConfig = "iproto"
-	hashConfigSymmetric hashConfig = "symmetric"
-	hashConfigReverse   hashConfig = "reverse"
+	lbTypeECMP      lbType = "ECMP"
+	lbTypeMaglev    lbType = "Maglev"
+	lbTypeMaglevDSR lbType = "MaglevDSR"
 )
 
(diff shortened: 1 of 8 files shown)


calico-vpp-agent/services/service_handler.go
```diff
@@ -274,6 +282,7 @@ func (s *Server) sameServiceEntries(entries []types.CnatTranslateEntry, service
 
 func (s *Server) addServiceEntries(entries []types.CnatTranslateEntry, service *LocalService) {
 	for _, entry := range entries {
+		s.log.Infof("^^^^^^^^^^%+v", ip.IPFlowHashConfig(entry.HashConfig))
```
sknat, line 285: This looks like a troubleshooting log message 😄

calico-vpp-agent/services/service.go
```diff
@@ -31,6 +33,16 @@ const (
...
+var hashConfigMap = map[hashConfig]types.IPFlowHash{
+	hashConfigDstaddr:   types.FlowHashDstIP,
+	hashConfigDstport:   types.FlowHashDstPort,
+	hashConfigIproto:    types.FlowHashProto,
+	hashConfigReverse:   types.FlowHashReverse,
+	hashConfigSrcaddr:   types.FlowHashSrcIP,
+	hashConfigSrcport:   types.FlowHashSrcPort,
+	hashConfigSymmetric: types.FlowHashSymetric,
+}
+
```
sknat, line 45: Instead of doing an extra layer of mangling here, we could directly build the bitmap in `ParseServiceAnnotations()` as we already are doing a switch statement there

vpplink/types/cnat.go
```diff
@@ -34,6 +35,18 @@ const (
 	MaglevLB  = CnatLbType(cnat.CNAT_LB_TYPE_MAGLEV)
 )
 
+type CnatHashConfig uint8
```
sknat, line 38: Maybe worth adding a `String()` function on the `CnatHashConfig` type

sknat (commented):
Thanks, this looks nice, just a few comments

vpplink/types/cnat.go
sknat: Right, there is this, but the output will probably be really ugly...
e.g. usually `"IP_API_FLOW_HASH_SRC_IP|IP_API_FLOW_HASH_DST_IP|IP_API_FLOW_HASH_SRC_PORT|IP_API_FLOW_HASH_DST_PORT|IP_API_FLOW_HASH_PROTO"` ...

sknat (approved):

PR #641, opened 2023-10-11
cni: move the PBL to punt when VCL is on

sknat:
This patch moves the PBL to the pod's VRF as it was previously in the main VRF. This will enable pod to self communication respecting the port mapping defined in the annotation.

When VCL is enabled on the pod, we add the PBL in the punt table so that traffic is first looked up in VPP's session table, then hits the port lookup in the PBL. This allows using all three (memif,VCL,tuntap) for a single pod.


hedibouattour:
This is not creating pbl at all in case of vcl. Current code is mutually exclusive between pb land vcl, so I think we need a deeper refactoring.

sknat:
Thanks for the feedback @hedibouattour didn't pay attention to this one.
If you have cycles, feel free to push on top of this PR

sknat:
Closing in favor of #646

PR #646, opened 2023-10-17
Allow enabling memif and vcl

hedibouattour:
on top of #641 and #644


sknat (approved):
Thanks for fixing it !

PR #649, opened 2023-10-27
openshift: add podSecurity labels

onong:
Openshift has pod security admission enabled by default. Adding the required labels.


sknat (approved):
lgtm

PR #696, opened 2024-04-25
Fix trying to close vpplink on vpp segfault

sknat:
This patch removes the closing of the vpplink channel when VPP fails 
to start as if err != nil, vpp will always be nil.


onong (approved):
LGTM

PR #704, opened 2024-11-06
default to memifEnabled true

sknat:
This patch enables memif support by default. Memif support has been in beta for the last two years and has gone through extended testing. As this is a major differentiator for the Calico/VPP integration, 3.29.0 release is a good candidate for making it GA.

Please not that pods will still be required to export the relevant annotation for it to be actually configured by the CNI.


hedibouattour:
do we need to add it to the yamls?

hedibouattour (approved):

PR #715, opened 2025-03-17
yaml: increase buffers-per-numa for openshift scenario

onong:
Openshift scenario in CI fails to deploy due to buffer shortage; Double the buffers-per-numa.


sknat (approved):
lgmt, thanks

PR #719, opened 2025-03-21
Refactor the prometheus exporter

sknat:
This patch refactors the Prometheus exporter code.
This should prevent the occurrence of race conditions where interfaces are
deleted while the controlplane is running and indexes are off.

Additionally, this adds supports for displaying VPP side interface names and
properly tagging memifs (displaying the name of the abstract socket)

### DIFF
--- a/calico-vpp-agent/prometheus/prometheus.go
+++ b/calico-vpp-agent/prometheus/prometheus.go
@@ -65,7 +65,7 @@ func (s *Server) recordMetrics(t *tomb.Tomb) {
 		}
 	}()
 	ticker := time.NewTicker(*config.GetCalicoVppInitialConfig().PrometheusRecordMetricInterval)
-	for ;t.Alive() ; <-ticker.C {
+	for ; t.Alive(); <-ticker.C {
 		ifNames, dumpStats, _ := vpplink.GetInterfaceStats(s.sc)
 		for _, sta := range dumpStats {
 			if string(sta.Name) != "/if/names" {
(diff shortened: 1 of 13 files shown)


calico-vpp-agent/prometheus/prometheus.go
```diff
@@ -38,236 +37,372 @@ import (
...
+				delete(self.podInterfacesByKey, initialPod.Key())
 				if podSpec.TunTapSwIfIndex == vpplink.INVALID_SW_IF_INDEX {
-					delete(s.podInterfacesBySwifIndex, initialPod.MemifSwIfIndex)
+					delete(self.podInterfacesDetailsBySwifIndex, initialPod.MemifSwIfIndex)
 				} else {
-					delete(s.podInterfacesBySwifIndex, initialPod.TunTapSwIfIndex)
+					delete(self.podInterfacesDetailsBySwifIndex, initialPod.TunTapSwIfIndex)
 				}
-				s.lock.Unlock()
+				self.lock.Unlock()
```
sknat, line 386: Same thing here

calico-vpp-agent/prometheus/prometheus.go
```diff
@@ -38,236 +37,371 @@ import (
...
+	self.log.Infof("Serve() Prometheus exporter")
 	go func() {
 		for t.Alive() {
 			/* Note: we will only receive events we ask for when registering the chan */
-			evt := <-s.channel
+			evt := <-self.channel
 			switch evt.Type {
 			case common.PodAdded:
+				self.lock.Lock()
+				defer self.lock.Unlock()
```
sknat, line 343: `defer` will only run when you exit the function scope, so we either need to make the `case` statement a function or call `Unlock()` explicitly both in `continue` and at the end of the processing

calico-vpp-agent/prometheus/prometheus.go
```diff
@@ -38,236 +37,371 @@ import (
...
+						interfaceName: podSpec.InterfaceName,
+					}
 				}
-				s.podInterfacesByKey[podSpec.Key()] = *podSpec
-				s.lock.Unlock()
+				self.podInterfacesByKey[podSpec.Key()] = *podSpec
 			case common.PodDeleted:
-				s.lock.Lock()
+				self.lock.Lock()
+				defer self.lock.Unlock()
```
sknat, line 373: Same here

sknat (commented):
Lgtm !

onong (approved):

PR #722, opened 2025-04-09
Add auto corepining path

sknat:
This patch adds support in VPP for pinning workers to the CPU cores that are assigned by a static allocator.

When k8s is configured to run with a static allocator [0] and that VPP starts with requests.CPU equal to limits.CPU and both integer numbers.

having VPP configured with
```
cpu {
   main-core 0
   corelist-workers 2-3
   relative
}
```

Will make it so that numbers 0,2,3 do not refer to absolutes CPU IDs but to the 0th, 2nd and 3rd CPU of those allotted by the static core allocator in k8s.

[0] https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy

### DIFF
--- a/vpplink/generated/vpp_clone_current.sh
+++ b/vpplink/generated/vpp_clone_current.sh
@@ -96,6 +96,7 @@ function git_clone_cd_and_reset ()
 git_clone_cd_and_reset "$1" cfa0953251cbab435307baf3dcd249fd95afaf1f # misc: VPP 24.10 Release Notes
 
 git_cherry_pick refs/changes/26/34726/3 # 34726: interface: add buffer stats api | https://gerrit.fd.io/r/c/vpp/+/34726
+git_cherry_pick refs/changes/94/41094/24 # 41094: vlib: improve core pinning | https://gerrit.fd.io/r/c/vpp/+/41094
 
 # This is the commit which broke IPv6 from v3.28.0 onwards.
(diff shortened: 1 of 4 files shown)


hedibouattour:
The patch got upstreamed, shall we just upgrade our VPP version?

hedibouattour:
Ah forgot we actually use 24.10 here, so LGTM

vpplink/generated/vpp_clone_current.sh
```diff
@@ -96,6 +96,7 @@ function git_clone_cd_and_reset ()
 git_clone_cd_and_reset "$1" cfa0953251cbab435307baf3dcd249fd95afaf1f # misc: VPP 24.10 Release Notes
 
 git_cherry_pick refs/changes/26/34726/3 # 34726: interface: add buffer stats api | https://gerrit.fd.io/r/c/vpp/+/34726
+git_cherry_pick refs/changes/94/41094/20 # 41094: vlib: improve core pinning | https://gerrit.fd.io/r/c/vpp/+/41094
```
hedibouattour, line 99: maybe 24th patchset, and btw this got upstreamed

hedibouattour (changes requested):

hedibouattour:
And this needs a run of `go mod tidy`

vpplink/generated/vpp_clone_current.sh
sknat: Good point, updated

PR #725, opened 2025-04-22
Add and document vcl sidecar

hedibouattour:
This adds a sidecar that runs with Pods that use VCL. It allows to have all libraries and dependencies in a volume and mount it on your application pod. This also adds vcl config as a configmap. Images are built and pushed to the repo.


test/yaml/iperf3-vclsidecar/test.yaml
```diff
@@ -0,0 +1,125 @@
...
+          - name: libraries
+            mountPath: /libraries              
+        imagePullPolicy: Always
+        ports:
+        - containerPort: 5201
+      - name: vclsidecar
+        image: docker.io/calicovpp/vclsidecar:v3.29.0
+        command: ["/bin/sh", "-c"]
+        args:
+          - cp /usr/local/lib/* /libraries && sleep infinity
```
sknat, line 47: We can maybe simplify this using something like this [0]
which would prevent us from having to copy
```yaml
spec:
  containers:
  - name: shell
    command: ["sleep", "infinity"]
    image: debian
    volumeMounts:
    - name: volume
      mountPath: /volume
  volumes:
  - name: volume
    image:
      reference: quay.io/crio/artifact:v1
      pullPolicy: IfNotPresent
```
[0] https://kubernetes.io/docs/concepts/storage/volumes/#image

test/yaml/sidecar-vcl/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:22.04
```
sknat, line 1: If we do not need to copy, we can make this image be
```Dockerfile
FROM scratch
```

test/yaml/Makefile
```diff
@@ -34,6 +34,15 @@ iperf3-vcl-image:
...
 
+.PHONY: sidecar-vcl
+sidecar-vcl:
+	for f in libvcl_ldpreload.so libvppcom.so.24.10.0 libvlibmemoryclient.so.24.10.0 \
+	        libsvm.so.24.10.0 libvppinfra.so.24.10.0 libvlibapi.so.24.10.0 ; do \
+		cp ${VPP_DIR}/build-root/build-vpp-native/vpp/lib/x86_64-linux-gnu/$${f} ./sidecar-vcl/ ; \
+	done
+	cp ${VPP_DIR}/extras/libmemif/build/lib/libmemif.so ./sidecar-vcl/
+	docker build -t calicovpp/vclsidecar -f ./sidecar-vcl/Dockerfile ./sidecar-vcl
+
```
sknat, line 45: Can we make this part of the `make image` and `make push` directives of
`vpp-manager/Makefile` ? That way we have CI taking care of it.

Also we should probably not depend on the VPP version i.e. something like :
```make
	for pkg in vpp vpp-plugin-core vpp-plugin-dpdk libvppinfra vpp-dbg ; do \
		cp vpp_build/build-root/$$pkg_*.deb $(IMAGE_DIR) ; \
	done
```

test/yaml/iperf3-vclsidecar/test.yaml
```diff
@@ -0,0 +1,125 @@
...
+          emptyDir: {}
+      containers:
+      - name: iperf3-server
+        image: calicovpp/test-iperf3:v3.29.0
+        command: ["tail", "-f", "/dev/null"]
+        env:
+        - name: VCL_CONFIG
+          value: /vcl/vcl.conf
+        - name: LD_LIBRARY_PATH
+          value: /libraries
```
sknat, line 34: Maybe we should use a more common path like `/usr/local/lib/vpp/`

sknat (commented):
looks good ! Just a few comments

test/yaml/iperf3-vclsidecar/test.yaml
hedibouattour: This is actually supported in Kubernetes v1.33 and still in Beta, I don't think we can use it for now, especially for current customer :)

hedibouattour:
#745 is merged

PR #729, opened 2025-06-20
Upgrade golangcilint version to v2.1.6

sknat:
This patch bumps the golangci lint version to v2.1.6, upgrades
the .golangcilint.yml config file and fixes the raised issues
in the golang source. Doing so we also:
- Upgrade the builder images to ubuntu:22.04
- Add coverage reporting for unittests
- Refactor ci Makefiles
- Default to using `CGO_ENABLED=0` for all agents


calico-vpp-agent/cni/storage/storage.go
```diff
@@ -36,7 +36,7 @@ import (
 
 const (
 	CniServerStateFileVersion = 8  // Used to ensure compatibility wen we reload data
```
hedibouattour, line 38: I think we need to increment this.

sknat: Good catch ! updated

hedibouattour (approved):

PR #730, opened 2025-06-23
add makefile target to build VPP from master

villinadrian:
Added new targets and modified the vpp_clone_current.sh script so we can build and test latest VPP. Also added proxy env variables to some 'docker run' commands to stop containers from timing out in Cisco labs.


hedibouattour (approved):

sknat (approved):
lgtm, thanks

PR #737, opened 2025-07-08
Add IP neighbor 30s aeging

sknat:
This patch enables IP neighbor aeging in VPP, defaulting to a 30sec renewal. IP neighbors older than 30s will trigger an ARP request or be retired.

This patch also marks the static neighbors we require as static so that they are not subject to aeging.

This addresses an issue in Cloud environment where nodes or peer VMs are recreated, keeping the same IP but changing the associated MAC without triggering a gratuitous ARP. When this happens, VPP keeps the old MAC entry indefinitely and traffic to said node is blackholed.

This can be configured using
```
CALICOVPP_INITIAL_CONFIG={
  "ip4NeighborsMaxNumber": 50000,
  "ip6NeighborsMaxNumber": 50000,
  "ip4NeighborsMaxAge": 30,
  "ip6NeighborsMaxAge": 30
}
```


hedibouattour (approved):
Thanks, lgtm.
But on a side note aren't we missing a del neighbor call at the`UnroutePodInterface` and `UnroutePblPortsPodInterface` level for L2 pods? or is it done automatically ?

PR #740, opened 2025-07-09
Support v6 services

hedibouattour:


calico-vpp-agent/services/service_server.go
```diff
@@ -325,48 +348,43 @@ func (s *Server) configureSnat() (err error) {
...
-	}
-	value, found, err := s.serviceStore.GetByKey(key)
-	if err != nil {
-		s.log.Errorf("Error getting service %s: %v", key, err)
-		return nil
-	}
-	if !found {
-		s.log.Debugf("Service %s not found", key)
+func (s *Server) findMatchingService(ep *discoveryv1.EndpointSlice) *v1.Service {
+	labelSvcName, ok := ep.Labels["kubernetes.io/service-name"]
```
sknat, line 352: Couldn't we keep the 
```
cache.DeletionHandlingMetaNamespaceKeyFunc(ep)
value, found, err := s.serviceStore.GetByKey(key)
```
?

calico-vpp-agent/services/service_server.go
```diff
@@ -325,48 +348,43 @@ func (s *Server) configureSnat() (err error) {
...
-	if !found {
-		s.log.Debugf("Endpoint %s not found", key)
-		return nil
-	}
-	endpoints, ok := value.(*v1.Endpoints)
-	if !ok {
-		panic("s.serviceStore.GetByKey did not return value of type *v1.Service")
+func (s *Server) findMatchingEndpoint(service *v1.Service) []*discoveryv1.EndpointSlice {
+	epslices := []*discoveryv1.EndpointSlice{}
+	endpointslices := s.endpointStore.List()
```
sknat, line 373: Here I think we should store the endpoint slices by service name in the 
same fashion as Calico does it here [0] to avoid the walk.

[0] https://github.com/projectcalico/calico/blob/master/felix/serviceindex/serviceindex.go#L38

calico-vpp-agent/services/service_server.go
```diff
@@ -158,7 +159,7 @@ func (s *Server) resolveLocalServiceFromService(service *v1.Service) *LocalServi
 	return s.GetLocalService(service, ep)
 }
 
-func (s *Server) resolveLocalServiceFromEndpoints(ep *v1.Endpoints) *LocalService {
+func (s *Server) resolveLocalServiceFromOldEndpointSlice(ep *discoveryv1.EndpointSlice) *LocalService {
```
sknat, line 162: Do we really need to differentiate code paths for old and new endpointslices ?

sknat (commented):
Looks good ! Thanks for taking a stab at it.
A few notes inline !

calico-vpp-agent/services/service_server.go
hedibouattour: Not really, we can use a flag..

calico-vpp-agent/services/service_server.go
```diff
@@ -79,6 +80,9 @@ type Server struct {
 	nodeBGPSpec *common.LocalNodeSpec
 
 	serviceStateMap map[string]ServiceState
+	// cache of all endpoint slices, by service name
+	endpointSlicesByService     map[string]map[string]*discoveryv1.EndpointSlice
+	endpointSlicesByServiceLock sync.Mutex
```
sknat, line 85: Let's just keep a single lock (`s.lock`) for simplicity

calico-vpp-agent/services/service_server.go
```diff
@@ -227,57 +260,50 @@ func NewServiceServer(vpp *vpplink.VppLink, k8sclient *kubernetes.Clientset, log
...
-							panic(fmt.Sprintf("obj.(cache.DeletedFinalStateUnknown).Obj not a (*v1.Endpoints) %v", obj))
+							panic(fmt.Sprintf("obj.(cache.DeletedFinalStateUnknown).Obj not a (*v1.EndpointSlice) %v", obj))
 						}
-						server.deleteServiceByName(serviceID(&endpoints.ObjectMeta))
-					case *v1.Endpoints:
-						server.deleteServiceByName(serviceID(&value.ObjectMeta))
-					default:
-						log.Errorf("unknown type in endpoints deleteFunction %v", obj)
 					}
+					server.deleteServiceByName(serviceID(&slice.ObjectMeta))
```
sknat, line 304: I don't think we can remove this here.
Probably better to call again :
```
server.handleServiceEndpointEvent(oldLocal, newLocal)
```

Also we'll need to get the name of the corresponding service being deleted, for the `` case, we'll need to keep a reference around to `endpointSlices` keyed by name so that we can find the trailing object, and fetch its name.

calico-vpp-agent/services/service_server.go
```diff
@@ -337,12 +363,15 @@ func (s *Server) configureSnat() (err error) {
 	return nil
 }
 
-func (s *Server) findMatchingService(ep *v1.Endpoints) *v1.Service {
+func (s *Server) findMatchingService(ep *discoveryv1.EndpointSlice) *v1.Service {
```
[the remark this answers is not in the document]
hedibouattour: The labels was the first logic I used 😄

PR #743, opened 2025-07-15
Rename policy server into felix server

hedibouattour:


Makefile
```diff
@@ -1,5 +1,7 @@
 include common.mk
 
+export GOOS=linux
```
sknat, line 3: Maybe `export GOOS ?= linux`
also this probably belongs in common.mk
Did you hit any issues with this ? we should be able to make the agent (and importantly the unit tests) on darwin as well as on linux

multinet-monitor/Makefile
```diff
@@ -5,6 +5,8 @@ VERSION_FILE=version
 GENERATE_LOG_FILE=../vpplink/generated/generate.log
 ALSO_LATEST ?= n # If 'y' also tag images with :latest
 
+export GOOS=linux
```
sknat, line 8: Same having it in `common.mk`  we would not require this

sknat (commented):
lgtm, thanks ! just a minor comment

Makefile
hedibouattour: This one is necessary for the go-lint to work, but yes putting it in common is enough for all.

PR #745, opened 2025-07-17
Include VCL sidecar build in CI pipeline

hedibouattour:

PR #751, opened 2025-07-22
Fix makefile VPP_HASH cache invalidation

sknat:
This patch addresses the fact that the VPP_HASH computation did not contain the list of deb files included within the VPP build cache. As a consequence, expanding this list, e.g. when added extra deps breaks the build.

This patch addresses this by making the deb list being part of the VPP_HASH computation.

This also enables the tar.gz computation when running locally, that way if the vpp_clone_current.sh has not changed, we won't rebuild VPP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ vpp-manager/vpp_build/
 *.deb
 *.so*
 multinet-monitor/watcher
+*.tgz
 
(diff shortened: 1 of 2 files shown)


sknat:
This supplements https://github.com/projectcalico/vpp-dataplane/pull/745

hedibouattour (approved):

PR #757, opened 2025-07-24
agent: remove policies enabled debug flag

hedibouattour:
this flag is not for policies anymore
it actually controls felix updates at large
so if disabled the agent will never start
maybe we should separate policies updates and flag them


sknat (approved):
lgtm, it does not make sense to keep the ability to disable the policy server if this breaks the rest of the felix functionalities

PR #773, opened 2025-09-02
Add make-kind directive

sknat:
This patch adds a 'make kind' directive that spins up a kind cluster, 
installs Calico, CalicoVPP from a local build VPP and the agent.

### DIFF
--- a/test/kind/install_calico_vpp.sh
+++ b/test/kind/install_calico_vpp.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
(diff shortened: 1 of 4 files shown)


test/kind/install_calico_vpp.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+kubectl create -f https://raw.githubusercontent.com/projectcalico/calico/master/manifests/tigera-operator.yaml
+
+while [[ "$(kubectl api-resources --api-group=operator.tigera.io | grep Installation)" == "" ]]; do echo "waiting for Installation kubectl resource"; sleep 2; done
```
villinadrian, line 7: wouldn't this suffice? 
`kubectl -n tigera-operator wait --for=condition=Ready pod --all`

sknat: Good point, let me change this

PR #774, opened 2025-09-02
[refactor] Split CNI into watcher/handler under felix

sknat:
This patch splits the CNI watcher and handlers
in two pieces. The handling will be done in the main
'felix' goroutine, while the watching / grpc server
will live under watchers/ and not store or access agent
state.

The intent is to move away from a model with multiple servers
replicating state and communicating over a pubsub. This being
prone to race conditions, deadlocks, and not providing many
benefits as scale & asynchronicity will not be a constraint
on nodes with relatively small number of pods (~100) as is k8s
default.


aritrbas:
Rebased on latest `master` to resolve merge conflicts and applied some fixes to comply with the latest Felix API updates in `release/v3.31.0` and `release/v3.32.0` as well as the NPOL and CNAT changes in VPP.

PR #777, opened 2025-09-03
add vpp prefix to prometheus stats

aritrbas:
add `vpp_` prefix to prometheus stats


calico-vpp-agent/prometheus/prometheus.go
```diff
@@ -86,20 +86,20 @@ func NewPrometheusServer(vpp *vpplink.VppLink, log *logrus.Entry) *PrometheusSer
 func cleanVppIfStatName(vppStatName string) string {
 	vppStatName = strings.TrimPrefix(vppStatName, "/if/")
 	vppStatName = strings.ReplaceAll(vppStatName, "-", "_")
-	return vppStatName
+	return "vpp_" + vppStatName
```
sknat, line 89: Could we make this configurable ? i.e. with a field in `CalicoVppInitialConfig` ?
We could default to something like `cni.projectcalico.vpp.` so that it has cni, calico and vpp in the name

sknat (commented):
Thanks @aritrbas !
Maybe we should add a config knob for it thought

sknat (approved):
thanks ! lgtm

PR #779, opened 2025-09-03
add prometheus integration tests

aritrbas:
Added integration tests for the Prometheus exporter feature.
It starts VPP in docker, then adds fake containers and make sure that the stats are properly reported to Prometheus.
It verifies that interface stats are exported per workers for each interface as well as TCP4/TCP6/session stats from VPP.

```
Usage: sudo INTEGRATION_TEST=1 VPP_IMAGE=calicovpp/vpp:latest VPP_BINARY=/usr/bin/vpp go test -v ./prometheus_test.go
```


sknat (approved):
lgtm ! Thanks

PR #833, opened 2025-11-14
doc/install: installation pre-requisites

onong:
When Calico/VPP is deployed, one of the things that happens is that the uplink interface vanishes from the root network namespace. Depending on the uplink driver used, it is either moved to a different network namespace or it vanishes completely from the kernel/os, and it gets replaced with a tap interface with the same name, mac address and other attributes. This Houdini Act may confuse the likes of `systemd-udevd`, `systemd-networkd` and `NetworkManager` and produce undesirable effects impacting the deployment and/or the functioning of the cluster.

Therefore, for successful deployment and functioning, Calico/VPP may require some changes in the configuration of system services pertaining to networking:
- udevd
- systemd-networkd
- NetworkManager

Adding documentation describing the issues and what needs to be done to circumvent them.

### DIFF
--- a/docs/install/NetworkManager.md
+++ b/docs/install/NetworkManager.md
@@ -1,0 +1,3 @@
+# Weird DHCP/DNS issue - NetworkManager
+
+TODO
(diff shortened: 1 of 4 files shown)


sknat (approved):
lgtm, thanks

PR #853, opened 2025-12-05
[refactor] vpp-manager rework

sknat:
Structural refactor of `vpp-manager`. No new networking features; the goal is to remove hidden global state, give `vpp-manager` a testable lifecycle and unify configuration between `vpp-manager` and `calico-vpp-agent`.

### 1. `pkg/config` is now the single source of truth:
- VRF constants (`DefaultVRFIndex`, `PuntTableID`, `PodVRFIndex`) moved out of `pkg/calico-vpp-agent/common` so both binaries reference identical values.
- A single `CreateVppLink(socket, log)` / `CreateVppLinkInRetryLoop(socket, log, timeout, retry)` now lives in `pkg/config/utils.go` and used by both `vpp-manager` and `calico-vpp-agent`.

### 2. `vpp-manager` lifecycle via `VppRunner`:
- `cmd/vpp-manager/main.go` reduced to argument parsing and `VppRunner.Run(ctx)`.
- `pkg/vpp-manager/vpp_runner.go` owns the entire lifecycle: signal handling, uplink driver invocation, VPP startup, hook orchestration and cleanup.
- Central `defer` recovers panics, signals VPP and runs the `VPP_ERRORED` hook on any failure.

### 3. Native Go hooks replace `default_hook.sh`:
- `pkg/vpp-manager/hooks/network_manager_hook.go` implements all five hook points in Go.
- Precedence:
  1. User-provided `CALICOVPP_HOOK_*` script (highest)
  2. Native Go hook when `CALICOVPP_ENABLE_NETWORK_MANAGER_HOOK=true` (default)
  3. Otherwise no-op
- Removes a bash dependency from the container image and replaces an indiscriminate `systemd-networkd`/`NetworkManager` restart with targeted Go logic.

### 4. Deterministic uplink driver selection:
- The legacy `SupportedUplinkDrivers` auto-detect retry loop is removed. The configured driver is invoked directly.
- `uplink/default.go::CreateMainVppInterface` now resolves the interface via `SearchInterfaceWithTag` from the VPP startup config.

### 5. Deterministic interface iteration:
- Operational loops in `vpp-manager` now iterate `InterfacesByID` (ordered slice) instead of `Interfaces` (Go map) so template rendering, driver `UpdateVppConfigFile`/`PreconfigureLinux`/`RestoreLinux`, and hook per-interface processing produce identical results run-to-run.

## Behavior Changes:

- **Uplink driver auto-detection retry is gone.** If the configured driver cannot bind the NIC, `vpp-manager` exits and is restarted by the DaemonSet rather than silently iterating `virtio → rdma → vmxnet3 → af_xdp → af_packet`. Deployments that relied on auto-detection must set `CALICOVPP_INTERFACES[*].vppDriver` (or the legacy `CALICOVPP_NATIVE_DRIVER`) explicitly.
- **`IfConfigSavePath` save/reload fallback removed.** The mechanism that persisted Linux interface state to a file and reloaded it when discovery failed at restart has been removed. `NewLinuxInterfaceState` now either succeeds or returns an error. Users who relied on `ifConfigSavePath` should remove that field from their `CalicoVppInitialConfig`; failure to discover the configured uplink will now cause `vpp-manager` to fail-fast and the DaemonSet to restart it (avoids booting against stale state).
- **`default_hook.sh` embedded fallback removed.** When `CALICOVPP_ENABLE_NETWORK_MANAGER_HOOK=false` and no `CALICOVPP_HOOK_*` user script is configured, the hook becomes a no-op (previously it ran the embedded bash script which always restarted `systemd-networkd`/`NetworkManager`). The `pkg/config/default_hook.sh` file is retained as a reference for users authoring their own hook scripts.
- **`vpp-manager` container image no longer requires `bash`** for hook execution.


config/config.go
```diff
@@ -619,9 +617,35 @@ const (
 	VfioUnsafeNoIommuModeDISABLED UnsafeNoIommuMode = "disabled"
 )
 
+type UplinkDriver interface {
+	PreconfigureLinux() error
+	CreateMainVppInterface(vpp *vpplink.VppLink, vppPid int, uplinkSpec *UplinkInterfaceSpec) error
+	RestoreLinux()
+	IsSupported(warn bool) bool
```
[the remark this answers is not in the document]
sknat: Very good point :open_mouth: 
Fixed it !

vpp-manager/startup/startup.go
```diff
@@ -99,40 +108,50 @@ func NewVppManagerParams() *config.VppManagerParams {
...
 	}
 
+	for _, intf := range params.Interfaces {
+		uplinkState, err := loadInterfaceConfigFromLinux(intf.Spec)
+		if err != nil {
+			log.Panicf("Could not load config from linux (%v)", err)
+		}
+		intf.State = uplinkState
+		intf.Driver = uplink.NewUplinkDriver(
+			intf.Driver.GetName(),
```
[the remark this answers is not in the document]
sknat: Agreed, this was poor implementation on my end, addressed it in the newer version.
Still I'm not really happy with the overall structure that smells of circular dependencies...

vpp-manager/uplink/af_xdp.go
```diff
@@ -149,19 +149,20 @@ func (d *AFXDPDriver) CreateMainVppInterface(vpp *vpplink.VppLink, vppPid int, u
...
-	d.params = params
-	d.spec = spec
-	return d
+func NewAFXDPDriver(params *config.VppManagerParams, intf *config.VppManagerInterface) *AFXDPDriver {
+	return &AFXDPDriver{
+		UplinkDriverData: UplinkDriverData{
+			name:   NativeDriverAfXdp,
+			params: params,
+			intf:   intf,
+		},
```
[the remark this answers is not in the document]
sknat: Agreed, same thing, went way too quick here, should be fixed now

pkg/vpp-manager/vpp_runner.go
```diff
@@ -17,961 +17,205 @@ package vppmanager
...
-					return []uint32{}, errors.Wrapf(err, "error add static neighbor for tap0 in VRF %d", vrfID)
+				v.log.Infof("Received signal %s", sig)
+				switch sig {
+				case syscall.SIGCHLD:
+					/* figure out pid of exited process */
+					wstatus := syscall.WaitStatus(0)
+					pid, err := syscall.Wait4(-1, &wstatus, syscall.WNOHANG, nil)
+					if err != nil {
+						v.log.Errorf("Wait4 error: %v", err)
+						continue
```
aritrbas: We should release `v.vppRunningCond.L.Lock()` here.

pkg/vpp-manager/vpp_runner.go
```diff
@@ -17,961 +17,205 @@ package vppmanager
...
+					}
+					if pid != v.vppProcess.Pid {
+						v.log.Infof("Ignoring SIGCHLD for pid %d", pid)
+						continue
```
aritrbas: We should release `v.vppRunningCond.L.Lock()` here.

pkg/vpp-manager/vpp_runner.go
```diff
@@ -1009,161 +253,122 @@ func (v *VppRunner) runVpp() (err error) {
...
-		return fmt.Errorf("cannot connect to VPP after 10 tries")
-	}
-
 	err = v.doVppGlobalConfiguration()
 	if err != nil {
-		terminateVpp("Error configuring VPP: %v", err)
-		v.vpp.Close()
-		<-vppDeadChan
-		return errors.Wrap(err, "Error configuring VPP")
+		err = errors.Wrap(err, "Error configuring VPP")
```
aritrbas: We should return the error here --> `return errors.Wrap(err, "Error configuring VPP")`

pkg/vpp-manager/vpp_runner.go
aritrbas: Fixed.

pkg/vpp-manager/vpp_runner.go
aritrbas: Fixed.

pkg/vpp-manager/vpp_runner.go
aritrbas: Fixed.

PR #918, opened 2026-02-21
vpp-manager: defer uplink admin-up until full FIB and tap setup

aritrbas:
### Summary
Defer bringing the VPP uplink interface admin-up until the very end of `runVpp()` after the full FIB has been programmed and the host-side taps have been configured. This eliminates a startup window during which RA processing and incomplete FIB state could cause incorrect forwarding or pollute the FIB.

### RCA
`runVpp()` previously set the uplink admin-up immediately after `CreateMainVppInterface()` and only then proceeded to configure RA suppression, CNAT features and VRF assignment. During that window:
- VPP could process RAs arriving on the wire and install unwanted default routes from upstream routers into the FIB.
- VPP could acquire SLAAC addresses on the uplink that it should not have.
- A less-specific route (e.g. a default GW installed first) could be hit before a more-specific connected prefix was installed, causing the first packets to be forwarded the wrong way.

### Fix
- Bring uplink admin-up at the end of `runVpp()`. Bringing the uplink up just before the tap means the moment the host sees its tap go up, the wire side is already live. This ensures early host-originated traffic (DHCPv6 SOLICIT, NS/NA, BGP keepalives, etc.) does not hit an admin-down uplink and get dropped in the FIB. VPP auto-generates the v6 link-local address at admin-up. Since `SetInterfaceVRF()` has already run, the link-local address is installed in the correct VRF.
- The `Retry(2*time.Second, 10, ...)` wrapper around `InterfaceAdminUp` is also dropped. By the time admin-up runs, many VPP API calls have already succeeded, so the binary API is demonstrably ready - the retry's old role as a VPP-readiness probe no longer applies.

#### Before
`CreateMainVppInterface` → `EnableInterfaceIP6` → **`InterfaceAdminUp` (uplinks)** → `configureVppUplinkInterface` (RA suppress, CNAT, VRF, addresses, routes) → tap setup → `configureLinuxTap` → `InterfaceAdminUp` (taps)
#### After
`CreateMainVppInterface` → `EnableInterfaceIP6` → `configureVppUplinkInterface` (RA suppress, CNAT, VRF, addresses, routes) → tap setup → `configureLinuxTap` → **`InterfaceAdminUp` (uplinks)** → `InterfaceAdminUp` (taps)

### DIFF
--- a/vpp-manager/vpp_runner.go
+++ b/vpp-manager/vpp_runner.go
@@ -1014,15 +1014,6 @@ func (v *VppRunner) runVpp() (err error) {
 			return errors.Wrap(err, "Error creating uplink interface")
 		}
 
-		// Data interface configuration
-		err = v.vpp.Retry(2*time.Second, 10, v.vpp.InterfaceAdminUp, v.params.UplinksSpecs[idx].SwIfIndex)
-		if err != nil {
-			terminateVpp("Error setting uplink interface up: %v", err)
-			v.vpp.Close()
-			<-vppDeadChan
-			return errors.Wrap(err, "Error setting uplink interface up")
-		}
-
 		err = v.configureVppUplinkInterface(v.uplinkDriver[idx], v.conf[idx], v.params.UplinksSpecs[idx])
 
 		if err != nil {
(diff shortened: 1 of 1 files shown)


vpp-manager/vpp_runner.go
```diff
@@ -611,6 +611,13 @@ func (v *VppRunner) configureVppUplinkInterface(
 		}
 	}
 
+	// Keep the interface admin-down until RA suppression and VRF binding are
+	// complete, then bring it up before programming IPv6/L3 state.
+	err = v.vpp.Retry(2*time.Second, 10, v.vpp.InterfaceAdminUp, ifSpec.SwIfIndex)
```
sknat: I think we should be able to skip the `Retry` here, iirc this was put in place to make sure
we would retry connecting if VPP was not ready as adminup was our first api call.

vpp-manager/vpp_runner.go
```diff
@@ -611,6 +611,13 @@ func (v *VppRunner) configureVppUplinkInterface(
...
+	if err != nil {
+		return errors.Wrap(err, "Error setting uplink interface up")
+	}
+
```
sknat: On another note, I think arguably we should be doing the admin-up last,
i.e. even after issuing `vpp.AddInterfaceAddress()` and `vpp.RouteAdd()` 
otherwise there is a (small) chance that we route the first packets the wrong
way if we happen to configure the less specific routes first

sknat (commented):
thanks for taking a stab at this !
A few comments inline

vpp-manager/vpp_runner.go
aritrbas: Yes! Removed in the latest commit.

vpp-manager/vpp_runner.go
aritrbas: Agreed. Updated in the latest commit.

sknat (approved):
lgtm

PR #964, opened 2026-04-07
vpp-manager: fix DHCPv6 ip6tables workaround

aritrbas:
When VPP sits between a Linux host tap and a physical uplink, it acts as a transparent forwarder for host-originated traffic. However, the ip4/ip6-rewrite nodes decrement TTL/hop-limit on every forwarded packet, breaking protocols that transmit with `TTL=1` (e.g. DHCPv6, eBGP single-hop, etc.)
Previously this was worked around with an `ip6tables` rule, which was fragile and protocol-specific (DHCPv6 only).

Implemented a VPP feature-arc based TTL fixup:
* **VPP patch**: Registers `ip4-ttl-fixup` / `ip6-ttl-fixup` nodes on the `ip4-unicast` / `ip6-unicast` input feature arcs.
* **CalicoVPP**: Enables the feature on the host tap interface during setup, pointing to the uplink as the destination.

The node performs a FIB lookup to determine the output adjacency. When the adjacency points to the configured uplink, it marks the packet with the `VNET_BUFFER_F_LOCALLY_ORIGINATED` flag so the standard rewrite skips TTL decrement.

This approach ensures:
- Zero overhead for the dominant traffic paths (pod↔pod, pod↔external).
- Only host-originated tap traffic (low volume) traverses the extra node.

This PR supersedes #961


vpplink/generated/patches/0006-feature-arc-TTL-fixup.patch
```diff
@@ -0,0 +1,697 @@
...
+This approach confines overhead to the configured source interface
+only. Packets arriving on all other interfaces have zero overhead.
+
+Type: feature
+
+Change-Id: Ie508ea5ac321457f398d67c9e52b5e603559d768
+Signed-off-by: Aritra Basu <aritrbas@cisco.com>
+---
+ src/vnet/CMakeLists.txt      |   2 +
+ src/vnet/ip/ip_ttl_fixup.api |  31 ++
```
sknat: Could we maybe make this its own plugin e.g. `src/plugins/ip_ttl_fixup/xxx` ?
Also, ideally as this is intended to be a private plugin moving forward, could
we instead of a .patch file make a 
```
vpplink/generated/private_plugins/ip_ttl_fixup/
```
directory and have the bare files in it.

In `vpp_clone_current.sh` instead of `git_apply_private` we could create a
new function that would be `copy_private_plugin` and just `rm -r ${VPP}/src/plugins/{name} && cp -r vpplink/generated/private_plugins/{name} ${VPP}/src/plugins/{name}`

sknat (commented):
Code looks neat ! Thanks for taking a stab at it !
Just a minor form-factor comment !

vpplink/generated/patches/0006-feature-arc-TTL-fixup.patch
aritrbas: Ack. Fixed in the latest commit.

sknat (approved):
lgtm, thanks !

Issue #962, opened 2026-03-28
SRPolicyNLRI missing Color field breaks dual-stack SRv6 (RFC 9256 §2.1 violation)


ryskn:
## Summary

In `calico-vpp-agent/common/common.go`, `MakePathSRv6Tunnel()` constructs `SRPolicyNLRI` without setting the `Color` field:

```go
nlrisr, err := apb.New(&bgpapi.SRPolicyNLRI{
    Length:   192,
    Endpoint: nodeIP,
})
```

Per [RFC 9256 §2.1](https://www.rfc-editor.org/rfc/rfc9256#section-2.1), an SR Policy is identified by the tuple `<Headend, Color, Endpoint>`, and Color is defined as:

> "an unsigned **non-zero** 32-bit integer value"

The current code leaves Color at its zero value, which:
1. Violates the RFC's "non-zero" requirement
2. Causes DT4 and DT6 SR Policies for the same endpoint to collide in GoBGP's RIB (same NLRI key), so only one survives

## Details

`MakePathSRv6Tunnel()` receives a `trafficType` parameter (4 for IPv4/DT4, 6 for IPv6/DT6), but this value is only used to set `SRv6EndPointBehavior` — it is never passed to the NLRI:

```go
func MakePathSRv6Tunnel(localSid net.IP, bSid net.IP, nodeIPv6 net.IP, trafficType int, ...) {
    if trafficType == 4 {
        epbs.Behavior = bgpapi.SRv6Behavior_END_DT4
    } else {
        epbs.Behavior = bgpapi.SRv6Behavior_END_DT6
    }

    nlrisr, err := apb.New(&bgpapi.SRPolicyNLRI{
        Length:   192,
        // Color is not set — defaults to 0 for both DT4 and DT6
        Endpoint: nodeIP,
    })
}
```

The codebase clearly intends dual-stack support. In `watchers/srv6_localsid_watcher.go`:

```go
switch localsid.Behavior {
case types.SrBehaviorDT4: // trafficType = 4
case types.SrBehaviorDT6: // trafficType = 6
}
```

And in `connectivity/srv6.go`, both `endDt4Exist` and `endDt6Exist` are tracked separately.

## Impact

In a dual-stack cluster (IPv4 + IPv6 pod CIDRs) with SRv6 enabled:
- `MakePathSRv6Tunnel()` is called twice per node — once for DT4, once for DT6
[...]

PR #967, opened 2026-04-08
srv6: fix dual-stack NLRI collision

aritrbas:
## Summary:
Fix RFC 9256 §2.1 violation in `MakePathSRv6Tunnel()` where `SRPolicyNLRI.Color` defaults to zero, breaking dual-stack SRv6 (DT4 + DT6) deployments.

## Root Cause:
`MakePathSRv6Tunnel()` constructs `SRPolicyNLRI` without setting the `Color` field:
```go
nlrisr, err := apb.New(&bgpapi.SRPolicyNLRI{
    Length:   192,
    Endpoint: nodeIP,
})
```
Per RFC 9256 §2.1, an SR Policy is identified by `<Headend, Color, Endpoint>` and `Color` must be "an unsigned non-zero 32-bit integer value". With `Color` defaulting to 0, we have:
- **RFC violation**: `Color=0` is explicitly prohibited.
- **NLRI collision**: Both DT4 and DT6 produce identical NLRI keys for the same node.
- **Data loss**: GoBGP's `AddPath` treats them as the same policy - the second call silently overwrites the first, leaving only one of IPv4/IPv6 forwarding working per node.

## Fix:
- Set `Color: uint32(trafficType)` where `trafficType` is 4 (DT4) or 6 (DT6). This ensures each `<Color, Endpoint>` tuple is unique and `Color` is non-zero.
- Also added unit tests to verify non-zero `Color` & DT4/DT6 NLRI uniqueness.

This fixes the issue identified in https://github.com/projectcalico/vpp-dataplane/issues/962

### DIFF
--- a/calico-vpp-agent/common/common.go
+++ b/calico-vpp-agent/common/common.go
@@ -383,6 +383,7 @@ func MakePathSRv6Tunnel(localSid net.IP, bSid net.IP, nodeIPv6 net.IP, trafficTy
 
 	nlrisr, err := apb.New(&bgpapi.SRPolicyNLRI{
 		Length:   192,
+		Color:    uint32(trafficType),
 		Endpoint: nodeIP,
 	})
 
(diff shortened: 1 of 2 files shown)


sknat (approved):
thanks a lot @aritrbas !

PR #973, opened 2026-04-18
fix: correct BSID parsing and SR Policy SAFI matching in SRv6

### LINKED ISSUE #972: SRv6: BGP path processing fails to install SR Policy correctly
## Summary
Three independent issues in the SRv6 BGP path processing pipeline
prevent SR Policy from being installed correctly in VPP. Combined, they
result in complete SRv6 forwarding failure without clear error messages.

### 1. BSID parsing silently fails
`getSRPolicy()` in `calico-vpp-agent/routing/bgp_watcher.go` unmarshals
the inner TLV into a raw `anypb.Any`, then tries to unmarshal that into
`SRBindingSID`:

```go
srbsids := &anypb.Any{}
if err := innerTlv.UnmarshalTo(srbsids); err == nil {
    if err := srbsids.UnmarshalTo(srv6bsid); err != nil {
        return nil, nil, nil, err
    }
}
```

The first `UnmarshalTo` fails with a type URL mismatch
(`apipb.TunnelEncapSubTLVSRBindingSID` vs. `google.protobuf.Any`), so
the block is silently skipped. The returned error is never propagated
because the work is gated on `if err == nil`. As a result,
`srv6bsid.Sid` stays `nil`, and VPP ends up with an SR Policy whose
BSID is all zeros. No traffic gets steered, and no error is logged.

### 2. SAFI check never matches
In `startBGPMonitoring()`, the SR Policy discriminator is:

```go
if ... && path.GetFamily() == &common.BgpFamilySRv6IPv6 {
```

Both operands are `*bgpapi.Family` pointers to distinct objects, so
`==` performs pointer comparison and is always `false`. SR Policy
paths therefore bypass `injectSRv6Policy()` entirely and fall through
to `injectRoute()`.

### 3. `injectRoute()` has no path for `SRPolicyNLRI`
`injectRoute()` only tries to unmarshal the NLRI into
`IPAddressPrefix` or `LabeledVPNIPAddressPrefix`. Any SR Policy
paths that reach it are discarded with:

```
cannot handle Nlri: type_url:".../apipb.SRPolicyNLRI" value:"..."
```

## Impact
- SRv6 dataplane does not forward any pod traffic.
- Bug 1 (BSID) is particularly hard to diagnose: no error is logged,
  the SR Policy appears to install successfully, and `show sr policies`
  in VPP may list a policy with BSID `::`.

## Proposed Fix
1. Unmarshal the inner TLV into `TunnelEncapSubTLVSRBindingSID` and
   then unmarshal its `.Bsid` Any into `SRBindingSID` (with a nil
   guard on the optional `Bsid` field).
2. Replace the pointer comparison with an enum-value SAFI check:
   `path.GetFamily().GetSafi() == bgpapi.Family_SAFI_SR_POLICY`.
3. In `injectRoute()`, detect `SRPolicyNLRI` and redirect to
   `injectSRv6Policy()` as a defensive fallback. This also helps
   catch any future regression of the SAFI check.

ryskn:
## Summary
Fix three independent issues in SRv6 BGP path processing that together
prevent SR Policy from being installed correctly in VPP.

## Root Cause

**1. BSID parsing silently fails**
`getSRPolicy()` unmarshals the inner TLV into a raw `anypb.Any` and
then tries to unmarshal that into `SRBindingSID`. The first
`UnmarshalTo` fails with a type URL mismatch
(`TunnelEncapSubTLVSRBindingSID` vs. `google.protobuf.Any`), the
block is silently skipped, and `srv6bsid.Sid` stays `nil`. VPP ends
up with an SR Policy whose BSID is all zeros, so no traffic is
steered.

**2. SAFI check never matches**
`startBGPMonitoring()` compares `path.GetFamily() == &common.BgpFamilySRv6IPv6`.
Since both sides are pointers to distinct objects, the comparison is
always `false`. SR Policy paths bypass `injectSRv6Policy()` entirely
and fall through to `injectRoute()`, which emits `cannot handle Nlri`
and drops them.

**3. `injectRoute` has no path for `SRPolicyNLRI`**
Even if the SAFI check were to be bypassed for any reason,
`injectRoute()` only recognizes `IPAddressPrefix` and
`LabeledVPNIPAddressPrefix`, so SR Policy paths are lost.

## Fix
1. Unmarshal through `TunnelEncapSubTLVSRBindingSID` and its inner
   `.Bsid` Any, with a nil guard for the optional `Bsid` field.
2. Compare `path.GetFamily().GetSafi() == Family_SAFI_SR_POLICY`
   (enum-value compare, AFI-agnostic).
3. In `injectRoute()`, detect `SRPolicyNLRI` and redirect to
   `injectSRv6Policy()` as a defensive fallback.

Fixes #972

### DIFF
--- a/calico-vpp-agent/routing/bgp_watcher.go
+++ b/calico-vpp-agent/routing/bgp_watcher.go
@@ -22,7 +22,6 @@ import (
 
 	bgpapi "github.com/osrg/gobgp/v3/api"
 	"github.com/pkg/errors"
-	"google.golang.org/protobuf/types/known/anypb"
 	"gopkg.in/tomb.v2"
 
 	calicov3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
@@ -89,6 +88,12 @@ func (s *Server) injectRoute(path *bgpapi.Path) error {
 			}
 			vpn = true
 		} else {
+			// Fallback: handle SRPolicyNLRI that may not have been caught by the SAFI check
+			srnrli := &bgpapi.SRPolicyNLRI{}
+			if err := path.Nlri.UnmarshalTo(srnrli); err == nil {
+				s.log.Infof("injectRoute: SRPolicyNLRI fallback, redirecting to injectSRv6Policy")
+				return s.injectSRv6Policy(path)
+			}
 			return fmt.Errorf("cannot handle Nlri: %+v", path.Nlri)
 		}
 	}
@@ -147,13 +152,14 @@ func (s *Server) getSRPolicy(path *bgpapi.Path) (srv6Policy *types.SrPolicy, srv
 						}
 					}
 					// search for TunnelEncapSubTLVSRBindingSID
-					srbsids := &anypb.Any{}
-					if err := innerTlv.UnmarshalTo(srbsids); err == nil {
+					subTLVBsid := &bgpapi.TunnelEncapSubTLVSRBindingSID{}
+					if err := innerTlv.UnmarshalTo(subTLVBsid); err == nil {
 						s.log.Debugf("getSRPolicy TunnelEncapSubTLVSRBindingSID")
-						if err := srbsids.UnmarshalTo(srv6bsid); err != nil {
-							return nil, nil, nil, err
+						if subTLVBsid.Bsid != nil {
+							if err := subTLVBsid.Bsid.UnmarshalTo(srv6bsid); err != nil {
+								return nil, nil, nil, err
+							}
 						}
-
 					}
 
 					// search for TunnelEncapSubTLVSRPriority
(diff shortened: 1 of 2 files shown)


calico-vpp-agent/routing/bgp_watcher.go
```diff
@@ -89,6 +88,12 @@ func (s *Server) injectRoute(path *bgpapi.Path) error {
 			}
 			vpn = true
 		} else {
+			// Fallback: handle SRPolicyNLRI that may not have been caught by the SAFI check
+			srnrli := &bgpapi.SRPolicyNLRI{}
+			if err := path.Nlri.UnmarshalTo(srnrli); err == nil {
+				s.log.Infof("injectRoute: SRPolicyNLRI fallback, redirecting to injectSRv6Policy")
```
sknat: nit: this should probably be `Debug`

calico-vpp-agent/routing/bgp_watcher.go
```diff
@@ -193,11 +199,14 @@ func (s *Server) getSRPolicy(path *bgpapi.Path) (srv6Policy *types.SrPolicy, srv
 }
 
 func (s *Server) injectSRv6Policy(path *bgpapi.Path) error {
+	s.log.Infof("injectSRv6Policy called")
```
sknat: Same thing here, logging in debug would probably be better

calico-vpp-agent/routing/bgp_watcher.go
```diff
@@ -193,11 +199,14 @@ func (s *Server) getSRPolicy(path *bgpapi.Path) (srv6Policy *types.SrPolicy, srv
...
 	_, srv6tunnel, srnrli, err := s.getSRPolicy(path)
 
 	if err != nil {
 		return errors.Wrap(err, "error injectSRv6Policy")
 	}
+	s.log.Infof("injectSRv6Policy: endpoint=%s, dst=%s, behavior=%d, bsid=%s, sids=%v",
```
sknat: And same here

sknat (commented):
Thanks for the PR, the core of the logic looks good to me,
just a nitpick on the log level, we run with info level by default
so debug is better for the injectSRv6Policy notice

ryskn:
Thanks for the review! You're right, these are noisy on the normal SRv6 path — moved all three to Debugf and force-pushed to the branch.

sknat (approved):
lgtm, thanks

PR #974, opened 2026-04-21
[v3.29] vpp-manager: move HookVppRunning after tap LL reconciliation

aritrbas:
### Problem
In some boots, DHCPv6 `SOLICIT` continues using the tap auto-generated LL addr even after it was replaced with the physical uplink LL addr in `configureIPv6LinkLocal()`. This causes source LL addr mismatch and intermittent DHCPv6 failures.

### Root Cause
Earlier, `HookVppRunning` (which restarts `systemd-networkd`) was executed before Linux tap LL addr reconciliation. Because of this, `systemd-networkd` could intermittently initialize DHCPv6 against the tap auto-generated LL addr due to a race in the `systemd` code and then keep that state even after the LL addr reconciliation resulting in incorrect source LL addr being used in DHCPv6 `SOLICIT` leading to intermittent DHCPv6 failures.

### Fix
Reorder startup in `runVpp()` so `HookVppRunning` runs after:
- `configureLinuxTap()` (tap link set UP, addresses/routes restored)
- `configureIPv6LinkLocal()` (tap auto-LL replaced with physical LL)

**Old startup order:** `HookVppRunning` --> `configureLinuxTap` --> `configureIPv6LinkLocal` --> `InterfaceAdminUp (VPP tap)`
**New startup order:** `configureLinuxTap` --> `configureIPv6LinkLocal` --> `HookVppRunning` --> `InterfaceAdminUp (VPP tap)`
 `systemd-networkd` restart now happens after LL reconciliation, so DHCPv6 state/socket initialization uses the final LL addr.

### DIFF
--- a/vpp-manager/vpp_runner.go
+++ b/vpp-manager/vpp_runner.go
@@ -755,11 +755,10 @@ func (v *VppRunner) configureVppUplinkInterface(
 		return errors.Wrap(err, "error configuring vpptap0 as host intf")
 	}
 
-	// Get Linux link info for the tap but do NOT bring it up yet.
-	// Linux tap configuration (link up, addresses, routes) is deferred
-	// to after the VPP_RUNNING hook so that udev rules restoring
-	// ID_NET_NAME_* properties are in place before systemd-networkd
-	// sees the interface and computes the DHCPv6 IAID.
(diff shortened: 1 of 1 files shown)


sknat (approved):
lgtm ! Thanks

PR #982, opened 2026-04-24
fix(cnat): make SNAT policy remove idempotent on ENOENT

ryskn:
## Problem

On an IPv6 single-stack cluster, `calico-vpp-node` crashes with:

```
cnatSnatPolicyAddDelIf 4 failed: VPPApiError: No such entry (-6)
```

The error propagates through the agent's tomb and tears the whole DaemonSet into `CrashLoopBackOff`.

### Reproduction

- kubeadm init with `--pod-network-cidr=fd20::/64 --service-cidr=fd30::/108`
- `Installation` CR with a single IPv6 `ipPools` entry
- Create any workload Pod; the agent crashes on the next `IpamConfChanged` event

## Root cause

`IpamConfChanged` reconciles SNAT for each pod interface × each IP family by calling `EnableDisableCnatSNAT(swIfIndex, isIP6, desiredState)` (`calico-vpp-agent/cni/cni_server.go:354`). The reconciliation is intentionally declarative:

```go
for _, podSpec := range s.podInterfaceMap {
    for _, swIfIndex := range []uint32{podSpec.LoopbackSwIfIndex, podSpec.TunTapSwIfIndex, podSpec.MemifSwIfIndex} {
        if swIfIndex != vpplink.InvalidID {
            for _, ipFamily := range vpplink.IPFamilies {
                err := s.vpp.EnableDisableCnatSNAT(swIfIndex, ipFamily.IsIP6, podSpec.NeedsSnat(...))
                ...
            }
        }
    }
}
```

For IPv6-only Pods only the v6 SNAT entry is ever registered on the interface, so the v4 state is "not present". When `NeedsSnat(v4) == false` and the agent calls `disable` on that interface, VPP replies with `No such entry (-6)` because there is nothing to remove. The agent treats it as a fatal error and `tomb.Dying` tears down every reconciliation goroutine.

In dual-stack clusters both families are always registered at Pod creation, so the remove path always hits an existing entry and this bug never surfaces.

## Fix

Make `cnatSnatPolicyAddDelPodInterface` tolerate `ENOENT` on the remove path: if the desired state ("not present") already matches the actual state, treat it as a no-op.

```go
if !isAdd && strings.Contains(err.Error(), "No such entry") {
    return nil
}
```

`add` continues to return errors so duplicate-add regressions remain visible.

## Why this shape

Proper state tracking on the agent side (remember which family was registered per interface and only toggle that) would be a larger refactor. Given the failure mode is just "agent over-reconciles in a way VPP dislikes", pushing the idempotency into the vpplink wrapper is the smallest safe change that keeps the reconcile loop alive.

### Before

All `calico-vpp-node` Pods in `CrashLoopBackOff` on an IPv6 single-stack cluster; agent logs end with:

```
Tomb exited with Error enabling/disabling ip6 snat:
cnatSnatPolicyAddDelIf 4 failed: VPPApiError: No such entry (-6)
```

### After

With this patch (plus #973 for SRv6 BGP parsing) applied:

```
$ kubectl -n calico-vpp-dataplane get pods
NAME                    READY   STATUS    RESTARTS   AGE
calico-vpp-node-j4lp7   2/2     Running   0          30m
calico-vpp-node-mkjp2   2/2     Running   0          30m
calico-vpp-node-psp7d   2/2     Running   0          30m

$ kubectl -n calico-vpp-dataplane exec <vpp-pod> -c vpp -- vppctl show sr policies
[0].- BSID: cafe::183 ...
[1].- BSID: cafe::c3  ...
```

Pod-to-Pod traffic over SRv6 working end-to-end on the test cluster.

## Related

- #973 (merged) — BSID / SAFI parsing fix for SRv6 BGP, which made SR Policy install work end-to-end. This PR addresses a different layer of the same `IPv6 single-stack` story: keeping the agent alive when it reconciles SNAT on Pod interfaces.

### DIFF
--- a/vpplink/cnat.go
+++ b/vpplink/cnat.go
@@ -18,6 +18,7 @@ package vpplink
 import (
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/projectcalico/vpp-dataplane/v3/vpplink/generated/bindings/cnat"
 	"github.com/projectcalico/vpp-dataplane/v3/vpplink/generated/bindings/interface_types"
@@ -165,6 +166,14 @@ func (v *VppLink) cnatSnatPolicyAddDelPodInterface(swIfIndex uint32, isAdd bool,
 		Table:     table,
 	})
 	if err != nil {
+		// On remove (isAdd=false), the agent iterates over all (interface, family)
+		// combinations from podInterfaceMap on every IpamConfChanged event. For an
+		// IPv6 single-stack cluster, IPv4 SNAT was never registered on Pod
+		// interfaces, so VPP replies "No such entry (-6)" when the agent tries
+		// to remove it. Swallow the ENOENT to keep the reconciliation loop alive.
+		if !isAdd && strings.Contains(err.Error(), "No such entry") {
+			return nil
+		}
 		return fmt.Errorf("cnatSnatPolicyAddDelIf %+v failed: %w", swIfIndex, err)
 	}
 	return nil


hedibouattour:
Not sure why this would happen, shouldn't CNAT_POLICY_INCLUDE_V4 be an empty table in this case ? can you share the output of `vppctl sh cn snat-policy` ?

ryskn:
Thanks @sknat @hedibouattour — VPP-side fix makes sense for keeping the API contract consistent, happy to look into it.

Sharing some state from a repro on my cluster (pre-patch image) as a data point:

```
Source NAT
  ip4: 192.168.1.102;0
  ip6: fd00:1::102;0

Excluded prefixes: (6 entries, node IPs on both families)

Included v4 interfaces:                 ← empty

Included v6 interfaces:                 ← populated
  loop0
  tun1
  loop1
  tun2

k8s pod interfaces:
  tap0, loop0, tun1, loop1, tun2
```

So `INCLUDE_V4` really is empty (matches your guess) and disable on any pod iface hits ENOENT. Same shape shows up on `cnat_snat_policy_add_del_exclude_pfx` in the agent tomb sequence:

```
... IpamConfChanged received
... Enable/Disable interface[3] SNAT
... Enable/Disable interface[4] SNAT
... Tomb function errored: cnatSnatPolicyAddDelIf 4 failed: VPPApiError: No such entry (-6)
... tomb Dying
... delete SNAT prefix failed: VPPApiError: No such entry (-6)
```

So the idempotency gap seems to be a family of cnat del APIs, not just `_add_del_if`.

Two questions before I head to VPP:

- Would you want all cnat del APIs with the "del-from-map" shape covered, or just `_add_del_if` for now?
- While the VPP change lands and flows into the calico-vpp bump, would you consider keeping this PR as a temporary guard in vpplink (easy to revert), or should we just leave the crash until then?

Happy to share the full agent / vpp container logs if useful, and happy to pick up the VPP side either way

hedibouattour:
Actually looking at vpp code, we should be hitting
```
  if (!is_add && !cnat_snat_policy_entry_is_init (cpe))
    return VNET_API_ERROR_NO_SUCH_ENTRY;
```
cnat snat policy isn't intialized.
Initialization should have happened if the function was called with is_add==true on some interface (any one).
So probably in your use case you're changing the ipam before any pod is ever created ?
I think that the right place to call it is at pool_get_zero time, right where the entry is first created, rather than deferring it to the first is_add call on add_del_if or add_pfx.
Happy to have you contribute to that and look at VPP if you want.
And regarding your second question, this should flow quite easily in the VPP code for now because we already cherry pick a cnat patch in `vpp_clone_current.sh`.

ryskn:
Thanks @hedibouattour, the pointer clears it up — yep, v6-only means INCLUDE_V4 never gets initialized via is_add=true. Moving init into pool_get_zero makes sense.

I'll take this up on the VPP side (covering both add_del_if and add_del_exclude_pfx) and push to gerrit. Will close this PR once the VPP change is in-flight 🙏

ryskn:
Pushed the VPP-side fix to gerrit: https://gerrit.fd.io/r/c/vpp/+/45604

Once it lands and we bump the cherry-pick in vpp_clone_current.sh, this PR can be closed. Will leave it open for now as tracking.

hedibouattour:
Thanks for the patch, lgtm, we can cherry-pick it in vpp_clone_current.sh and wait for it to be merged.

ryskn:
Thanks! Closing this in favor of the VPP-side fix (https://gerrit.fd.io/r/c/vpp/+/45604), which you'll be cherry-picking into vpp_clone_current.sh.

PR #988, opened 2026-04-28
Fix protobuf issues

hedibouattour:


sknat (approved):
lgtm, many thanks @hedibouattour

PR #990, opened 2026-04-28
[v3.29] Harden testing 3.29

hedibouattour:
adding cnat drain test doesn't make a lot of sense because non natted sessions don't use cnat in this version.
But it's still worth it because it tests trex+memif+PBL.


sknat (approved):
lgtm, thanks

PR #991, opened 2026-04-28
run run-main-tests on kind in CI

sknat:
This patch runs the cases.sh test cases in CI on a locally deployed kind cluster.


sknat:
Abandoning again as we rely on Codebuild for now which ships with a ~4.14 kernel which 
requires arguably too much tweaking to get kind up.

PR #993, opened 2026-04-29
build: add private plugins to vpp tarball hash

hedibouattour:


sknat (approved):
Wow, good catch, thanks !

PR #996, opened 2026-05-04
keep policies naming consistent

hedibouattour:
In calico, we have ingress vs egress policies, policies are expressed from the POV of the pods/host
In vpp, we use rx vs tx, from the POV of VPP interfaces. So we reverse intent (except for uplink where we use invertRxTx) This patch keeps naming consistent.

`InterfaceConfig` is expressed in Calico manner.
`func (v *VppLink) ConfigurePolicies` is where we pass from Calico to VPP
convention.


cnhe-vpp-gen (approved):
lgtm

sknat (approved):
lgtm

PR #998, opened 2026-05-05
vpp: update to latest version and calico to 3.32

hedibouattour:
This patch promotes the base VPP commit we build
from 72bcbcd01 - vlib: fix one-time barrier release callbacks
to da090867b - vppinfra: zero-initialize crypto contexts with const members
This allows us to upstream the cnat single lookup refactoring
that went in as [0]

This patch also freezes the Calico version the kind install
directive targets to v3.32 as Calico v3.32 was just released.

[0] https://gerrit.fd.io/r/c/vpp/+/43369

### DIFF
--- a/test/kind/Makefile
+++ b/test/kind/Makefile
@@ -1,9 +1,9 @@
 include ../../common.mk
 
 CPU_PINNING ?= false
-KIND_CALICO_VERSION ?= master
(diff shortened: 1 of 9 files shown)


sknat (approved):
lgtm

PR #1004, opened 2026-05-05
Backport new tests 3.31

aritrbas:
Since we have backported all IPv6/DHCPv6 and NPOL changes in `release/v3.31.0`, perhaps we should backport the tests as well like we did in `release/v3.29.0`.


sknat (approved):
lgtm

PR #1008, opened 2026-05-06
[3.29] Backport CrossSubnet Fix

aritrbas:
### Problem
`CrossSubnet` mode for VXLAN and IPIP is not honored when the local node's BGP address has a **host prefix** (`/128`). Instead, it falls back to `Always`, forming tunnels to all peers including those within the same subnet.

### RCA
`CrossSubnet` provider selection relies on local node CIDR containment checks (`nodeIPNet.Contains(peerNextHop)`). On environments where the uplink IPv6 address is discovered as a **host prefix** (`/128`), `node.Spec.BGP.IPv6Address` was advertised as `/128` which never contains a same-subnet peer address, so `CrossSubnet` effectively behaved like `Always`.

### Fix
- Moved IPv6 `/128` -> `/64` translation into `LinuxInterfaceState` address getters instead of applying it only at VPP interface programming time.
- `updateCalicoNode()` now advertises translated IPv6 (`/64`) through `GetNodeIP6()`, so Felix `nodeBGPSpec` receives a subnet prefix usable by `CrossSubnet` checks.
- Added `GetAddressesNoMaskTranslation()` and switched Linux address add/del and tap restore/programming paths to use raw addresses, preserving previous Linux-side behavior.

The patch is already present on `master`, `release/v3.32.0`, `release/v3.31.0`, and `release/v3.30.0`.

### DIFF
--- a/calico-vpp-agent/connectivity/connectivity_server.go
+++ b/calico-vpp-agent/connectivity/connectivity_server.go
@@ -120,6 +120,9 @@ func (s *ConnectivityServer) GetNodeIPs() (ip4 *net.IP, ip6 *net.IP) {
 }
 
 func (s *ConnectivityServer) GetNodeIPNet(isv6 bool) *net.IPNet {
+	if s.nodeBGPSpec == nil {
+		return nil
+	}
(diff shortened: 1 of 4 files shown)


sknat (approved):
lgtm, thanks

PR #1009, opened 2026-05-07
felix: replace existing IPSetUpdate members

aritrbas:
This fixes the `FelixPolicyChurnTest` failure in VPP `kube-test`.

### RCA:
`handleIpsetUpdate()` returns a fatal error when an `IPSetUpdate` arrives for an IP set ID that already exists in the local state:
```
return fmt.Errorf("received new ipset for ID %s that already exists", msg.GetId())
```

`IPSetUpdate` carries the complete current membership of a set and is a valid `replace` operation. This message legitimately arrives during Felix resync causing the agent to crash-loop on policy synchronisation.

Upstream Calico's `policysync` handles this as follows:
```
s, ok := p.ipSetsByID[id]
if !ok {
    s = newIPSet(update)
    p.ipSetsByID[id] = s
    return
}
s.replaceMembers(update)
```

We were only handling the initial-sync path (create → delta-update → remove) but did not account for the resync path where Felix re-sends the full set state via `IPSetUpdate` for sets that were already programmed.

### Fix:
Added `ReplaceMembers()` that guards against type mismatch between old and incoming set by using a generic `diffStringMap` helper to compute add/remove slices for each member type (`IP`, `IPPort`, `Net`). It calls existing `RemoveMembers()` and `AddMembers()` methods which handle both the VPP dataplane calls (when `apply=true`) and the in-memory map updates (`always`).

### DIFF
--- a/calico-vpp-agent/felix/messages.go
+++ b/calico-vpp-agent/felix/messages.go
@@ -68,7 +68,8 @@ func (s *Server) RecvMessage(conn net.Conn) (msg interface{}, err error) {
 	envelope := proto.ToDataplane{}
 	err = pb.Unmarshal(data, &envelope)
 	if err != nil {
-		return
+		s.log.Warnf("Skipping invalid message from felix-dataplane.sock: %v", err)
+		return nil, nil
 	}
 
 	switch payload := envelope.Payload.(type) {
--- a/calico-vpp-agent/felix/felix_server.go
+++ b/calico-vpp-agent/felix/felix_server.go
@@ -750,9 +750,14 @@ func (s *Server) handleIpsetUpdate(msg *proto.IPSetUpdate, pending bool) (err er
 		return errors.Wrap(err, "cannot process IPSetUpdate")
 	}
 	state := s.currentState(pending)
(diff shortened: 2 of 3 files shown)


calico-vpp-agent/felix/messages.go
```diff
@@ -68,6 +73,7 @@ func (s *Server) RecvMessage(conn net.Conn) (msg interface{}, err error) {
 	envelope := proto.ToDataplane{}
 	err = pb.Unmarshal(data, &envelope)
 	if err != nil {
+		err = invalidFelixMessageError{err: err}
```
sknat: Let's maybe log here and return `msg = nil` to avoid having to interpret the error in the caller

sknat (commented):
lgtm, good catch !
Just a nitpick inline

calico-vpp-agent/felix/messages.go
aritrbas: Ack. Fixed in the latest commit.

sknat (approved):
lgtm, thanks !

PR #1010, opened 2026-05-08
[v3.29] vpp-manager: prefer real default-route gateway in pickNextHopIP

aritrbas:
### RCA
After a VPP crash and recovery, there is a potential **race window** where the `calico-vpp-agent`'s route-watch goroutine `Go(routeWatcher.WatchRoutes)` re-installs cluster routes (service/pod CIDRs) onto the physical uplink before `RestoreLinux()` has removed them. Those routes use `192.168.1.254` (or equivalent) as their gateway - an address that only exists inside VPP's ARP proxy on the tap interface.

When `vpp-manager` restarts, `LoadInterfaceConfigFromLinux()` picks up the stale route and programs it into VPP through the physical uplink interface. VPP tries to ARP-resolve `192.168.1.254` (or equivalent) on the wire, fails, and marks the FIB entry `UNRESOLVED`. All traffic to service and pod CIDRs is dropped until the next full restart.

### Fix
`pickNextHopIP` now first looks for a default route (`Dst==0.0.0.0/0` or `Dst==::/0` or `Dst==nil`) with a gateway and uses it as the next-hop IP instead of the synthetic address. The real gateway is ARP-resolvable on the physical segment, so even if a stale cluster route leaks onto the physical uplink and is re-programmed into VPP it will resolve correctly rather than going `UNRESOLVED`. Here is how `pickNextHopIP()` works now:
1. find a default route with a gateway [new]
2. existing broadcast-minus-one heuristic fallback
3. existing directly-connected-route heuristic fallback

With this change, `fakeNextHopIP4` becomes `192.168.1.1` (or equivalent) (i.e. the real router) instead of `192.168.1.254`. If a stale cluster route for `192.168.1.1` (or equivalent) ends up on the physical uplink and gets re-programmed by VPP, VPP can ARP-resolve the real router - the route is no longer `UNRESOLVED` and traffic continues to flow!

This is a cherry-pick of #1011 from `master`.

### DIFF
--- a/vpp-manager/vpp_runner.go
+++ b/vpp-manager/vpp_runner.go
@@ -278,6 +278,14 @@ found:
 	return nil
 }
 
+func isDefaultRoute(dst *net.IPNet) bool {
+	if dst == nil {
+		return true
+	}
+	ones, bits := dst.Mask.Size()
+	return ones == 0 && (bits == 32 || bits == 128)
+}
+
 // pick a next hop to use for cluster routes (services, pod cidrs) in the address prefix
 func (v *VppRunner) pickNextHopIP(ifState config.LinuxInterfaceState) (fakeNextHopIP4, fakeNextHopIP6 net.IP) {
 	fakeNextHopIP4 = net.ParseIP("0.0.0.0")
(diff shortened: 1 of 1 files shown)


sknat (approved):
lgtm, thanks !

PR #1016, opened 2026-05-11
[v3.29] routerwatcher: provide link index to netlink

sknat:
This patch passes the link index to the route we craft for ippools we need to route through the cluster.

This makes sure that the route is added for the tap created by VPP and not the uplink in the event where VPP has restared.

Guarantee is provided by the fact that the kernel does not immediately reuse link indexes, and that they are scoped by netns. We should thus see only the uplinks and the taps in the main netns, and have no risk of conflict between uplink and tap.


calico-vpp-agent/watchers/uplink_route_watcher.go
aritrbas, line 213: Should we also add `LinkIndex: uplinkStatus.LinkIndex` here for the service CIDR routes?

aritrbas (approved):

calico-vpp-agent/watchers/uplink_route_watcher.go
sknat: Good point, done !

PR #1012, opened 2026-05-10
fix(bgp): bound-check segment count in getSRPolicy

ryskn:
## Summary

`calico-vpp-agent/routing/bgp_watcher.go::getSRPolicy` has **no bound check on the segment count**, so a malformed SR Policy NLRI from a BGP peer (segment list of 0 or 17+ entries) crashes the agent process with a runtime panic. This PR adds a small guard that returns an `error` instead of panicking on malformed input.

---

### 1. VPP API uses a fixed 16-SID array

VPP's `src/vnet/srv6/sr.api`:

```c
typedef srv6_sid_list {
  u8 num_sids;
  u32 weight;
  vl_api_ip6_address_t sids[16];   // <- fixed length 16
};
```

This is auto-generated as `vpplink/generated/bindings/sr/sr.ba.go::Srv6SidList.Sids [16]ip_types.IP6Address`. The `16` is not a Calico-VPP convention — it is the **VPP plugin's API contract**.

### 2. Two out-of-range accesses in `getSRPolicy`

```go
// bgp_watcher.go (before fix)
segments := []*bgpapi.SegmentTypeB{}     // unbounded slice, peer-supplied
// ... unmarshal TunnelEncap sub-TLVs and append ...

// (A) 17+ entries -> fixed-array OOR
policySidListsids := [16]ip_types.IP6Address{}
for i, segment := range segments {
    policySidListsids[i] = ...           // panics at i = 16
}

// (B) 0 entries -> empty-slice tail dereference (index -1)
srv6tunnel.Behavior = uint8(
    segments[len(segments)-1].           // len(segments) = 0 -> index -1
        GetEndpointBehaviorStructure().Behavior,
)
```

`startBGPMonitoring` has no `defer recover()`, so the panic propagates from the `WatchBGPPath` goroutine and **kills the entire agent process** — i.e. a remote DoS of the node's routing daemon.

---

## Panic logs captured on a real cluster master node

Stack traces from invoking `getSRPolicy` with crafted 17-segment / 0-segment input on `master.ryskn.k8s` (`Linux 6.12.0-124.52.3.el10_1.x86_64`, AlmaLinux 10.1):

### Case A: 17 segments — fixed-array OOR

```
panic: runtime error: index out of range [16] with length 16 [recovered, repanicked]

goroutine 14 [running]:
panic({0x21c53e0?, 0x611b48ce240?})
	/opt/homebrew/Cellar/go/1.26.2/libexec/src/runtime/panic.go:860 +0x13a
[...]

PR #1018, opened 2026-05-12
fix(bgp): handle SR Policy withdraw and validate endpoint behavior in getSRPolicy

ryskn:
## Summary

Follow-up to PR #1012 (c0b8d78e). Closes the remaining two `getSRPolicy()` paths in `calico-vpp-agent/routing/bgp_watcher.go`.

| Case | Trigger | Pre-fix behaviour | This PR |
|---|---|---|---|
| Oversized segments | Peer sends SR Policy with 17+ SegmentTypeB | out-of-range panic while writing the 17th SID | Already fixed by #1012 (bound check) |
| **Withdraw** | `MP_UNREACH_NLRI` for SR Policy (RFC 9012) — empty Pattrs | Pre-#1012: `segments[-1]` panic. Post-#1012: drop without emitting `SRv6PolicyDeleted`, VPP keeps stale policy. | `path.IsWithdraw` short-circuit so `injectSRv6Policy` can dispatch the deletion event |
| **Nil EBS** | Peer sends SegmentTypeB without `EndpointBehaviorStructure` | `nil.Behavior` SIGSEGV → agent process abort | Reject the SR Policy with an error when the last segment lacks endpoint behavior info |

## Why these matter

SRv6 Endpoint Behavior and SID Structure is **optional** for Segment Type B per RFC 9830 (and described in RFC 9256 as MAY be provided for validation at the headend). Conforming peers may legally advertise SR Policy NLRIs without it. The receive path must tolerate that encoding, not dereference an optional field that may be nil.

Withdraw handling is also required for normal SR Policy lifecycle operations such as policy deletion or BGP session teardown. Without this path, a previously installed VPP SR policy / steering entry can remain stale because `SRv6PolicyDeleted` is never published.

In local interop testing, `gobgpd`-generated SR Policy updates can omit `EndpointBehaviorStructure`, which is what surfaced the nil dereference here.

## Prior reproduction in the same function

A reproducer triggered the oversized-segments case (already fixed by #1012) and the agent aborted at `getSRPolicy()` (`bgp_watcher.go:187`, `runtime error: index out of range [16] with length 16`, exitCode 2). The withdraw and nil-EBS cases live in the same peer-input parsing path. The withdraw case can leave stale VPP state by dropping withdraws before `SRv6PolicyDeleted` is emitted, while the nil-EBS case can still abort the agent process through a nil optional field dereference. The fixes here make both paths return explicit errors or deletion events before any unsafe slice / pointer access.

## Diff summary

`calico-vpp-agent/routing/bgp_watcher.go`, +23 / -1:

1. Early `path.IsWithdraw` return immediately after NLRI unmarshal, with `srv6tunnel.Dst` populated so `injectSRv6Policy` can still send `SRv6PolicyDeleted`.
2. Nil check on `GetEndpointBehaviorStructure()` before `.Behavior` access, returning a descriptive error on nil.

No new dependencies, no API changes, no test fixture changes.

## Notes

Same function as PR #1012; comments match the surrounding style.

### DIFF
--- a/calico-vpp-agent/routing/bgp_watcher.go
+++ b/calico-vpp-agent/routing/bgp_watcher.go
@@ -137,6 +137,16 @@ func (s *Server) getSRPolicy(path *bgpapi.Path) (srv6Policy *types.SrPolicy, srv
 	}
 	srv6tunnel.Dst = net.IP(srnrli.Endpoint)
 
+	// SR Policy withdraws carry only the (distinguisher, color, endpoint)
+	// NLRI key; per RFC 9012 / draft-ietf-idr-segment-routing-te-policy the
+	// TunnelEncap path attribute is not required. Short-circuit so the
+	// caller (injectSRv6Policy) can dispatch SRv6PolicyDeleted instead of
+	// reaching the empty-segments check below and dropping the withdraw
+	// without emitting SRv6PolicyDeleted.
+	if path.IsWithdraw {
+		return nil, srv6tunnel, srnrli, nil
+	}
+
 	for _, pattr := range path.Pattrs {
 		if err := pattr.UnmarshalTo(tun); err == nil {
 			for _, tlv := range tun.Tlvs {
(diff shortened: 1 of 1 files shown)


calico-vpp-agent/routing/bgp_watcher.go
```diff
@@ -137,6 +137,16 @@ func (s *Server) getSRPolicy(path *bgpapi.Path) (srv6Policy *types.SrPolicy, srv
...
 	srv6tunnel.Dst = net.IP(srnrli.Endpoint)
 
+	// SR Policy withdraws carry only the (distinguisher, color, endpoint)
+	// NLRI key; per RFC 9012 / draft-ietf-idr-segment-routing-te-policy the
+	// TunnelEncap path attribute is not required. Short-circuit so the
+	// caller (injectSRv6Policy) can dispatch SRv6PolicyDeleted instead of
+	// reaching the empty-segments check below and dropping the withdraw
+	// without emitting SRv6PolicyDeleted.
+	if path.IsWithdraw {
+		return nil, srv6tunnel, srnrli, nil
```
aritrbas, line 147: small nit:

If `path.IsWithdraw` is `true`, `getSRPolicy` will return `srv6tunnel` with only `Dst` populated. The caller `injectSRv6Policy` will build a `NodeConnectivity` object with this partial tunnel and send a `SRv6PolicyDeleted` event, the handler for which in `connectivity_server.go` will call `UpdateSRv6Policy(old, true)` that eventually calls `DelConnectivity(cn)` which is a `no-op` in `srv6.go` → `FIXME SRv6 node connectivity removal not supported`. If `DelConnectivity` is ever implemented, it might encounter `nil` values and likely panic.

**But this should be good for now.**

aritrbas (approved):
lgtm, thanks !

calico-vpp-agent/routing/bgp_watcher.go
ryskn: Thanks for tracing that through.

Agreed — the partial `srv6tunnel` is safe today only because `SRv6Provider.DelConnectivity` is still the FIXME no-op.

For this PR, I’d like to keep the scope limited to avoiding the withdraw-time crash and emitting the deletion event for stale policies, so I’d appreciate it if we could merge it as-is.

When `DelConnectivity` gets a real implementation, I think we’ll need either a nil-safe walk on the consumer side or default population in the withdraw branch of `getSRPolicy` for fields such as `Bsid` / `Policy`. I’ll handle that part in a follow-up PR.

PR #1020, opened 2026-05-14
calicovppctl: prevent export from hanging on unhealthy nodes

aritrbas:
### Summary

This fixes `calicovppctl export` from hanging when one or more nodes are in a bad state and VPP CLI commands issued through `kubectl exec` appear stuck as the failed commands fall back to the default OS I/O timeout, which could be up to 60s per call.
The patch makes `calicovppctl` enforce its own deadlines around non-interactive subprocesses and ensures `export` degrade gracefully when VPP command collection is failing.

### Changes
- Added a `requestTimeout` to `KubeClient`, currently set to **5 seconds**.
- Added `--request-timeout=5s` to `kubectl exec` calls.
- Added a shared `runCommandWithTimeout` helper using `exec.CommandContext` and routed non-interactive `kubectl` subprocess execution through that helper.
- Added a timeout context to the SPDY file streaming path.
- Kept VPP CLI commands on the shorter 5s timeout path.
- Kept general Kubernetes API operations on the existing 15s timeout.
- Preserved command errors in export files instead of silently dropping them.
- Added per-node VPP timeout tracking during export. After 3 consecutive VPP CLI timeouts on a node, remaining VPP CLI dump files for that node are written with explicit skip markers.

### RCA

`calicovppctl export` runs many VPP dump commands per node. If the VPP CLI path on a node is wedged, waiting on every command can make export appear stuck and delay collection of the rest of the diagnostic data.

`kubectl --request-timeout` only bounds part of the request behavior. Since `calicovppctl` launches `kubectl` as a local child process, `calicovppctl` also needs to enforce a parent-side deadline. Using `exec.CommandContext` guarantees that
the parent process does not wait indefinitely for a hung subprocess.

The skip-after-consecutive-timeouts behavior keeps the archive useful while bounding the worst case. The `export` still records which commands timed out or were skipped, then continues collecting Kubernetes state and pod/log data where possible.

### DIFF
--- a/cmd/calicovppctl/main.go
+++ b/cmd/calicovppctl/main.go
@@ -69,8 +69,14 @@ const (
 	cmdVppShowRun = "show run"
 	cmdVppShowErr = "show err"
 
-	// Kubernetes client timeout
+	// Kubernetes client timeout (for k8s API calls)
 	kubeClientTimeout = 15 * time.Second
+
+	// Default timeout (for kubectl exec commands and streaming requests)
+	defaultRequestTimeout = 5 * time.Second
+
+	// Stop trying VPP CLI commands on a node after this many consecutive timeouts
+	maxConsecutiveVppCommandTimeouts = 3
 
 	// Number of times capture file transfer is retried on failure
 	maxTransferRetries = 3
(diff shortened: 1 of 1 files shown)


sknat (approved):
lgtm

PR #1021, opened 2026-05-19
Fix external BGP SR Policy parsing and install path

ryskn:
## Summary

Fix the SR Policy parser and the VPP install path so multi-Segment-List `TunnelEncapSubTLVSRSegmentList` advertisements are preserved end-to-end instead of being flat-concatenated into one strict SID chain, and close several panic exits from `getSRPolicy()`. Also bridge `BGPPeer.TTLSecurity` into gobgp so external eBGP peers can establish across VPP's host-eth0 ↔ tap L3 hop -- without that, the SR Policy fix cannot be validated end-to-end.

Prior fixes (`projectcalico/vpp-dataplane#1012`, `#1018`) plugged some panics piecemeal but left the root cause -- no per-Segment-List parse function -- in place.

## Scope of this PR

Two fixes that together make external-BGP-controller SR Policy distribution validatable end-to-end:

1. **SR Policy parser and VPP install** (`calico-vpp-agent/routing/bgp_watcher.go`, `vpplink/srv6.go`). A mix of three categories, not "RFC-violation fixes" wholesale:
   - **RFC 9256 alignment.** Stop flat-concatenating multiple `TunnelEncapSubTLVSRSegmentList` entries into one chain. Independent Segment Lists are the RFC's ECMP / weighted-candidate-path mechanism; flatten silently changes the advertised forwarding semantics.
   - **Defensive validation.** Close per-list panic paths (`EndpointBehaviorStructure` nil deref, over-limit segment count, empty list, garbage `Any`) so malformed advertisements no longer crash the agent. Plain bug fixes.
   - **Implementation-constraint rejects.** `SegmentTypeA` (SR-MPLS) presence and cross-list trailing-behavior disagreement are *not* RFC-forbidden. They are rejected because Calico-VPP's install layer is built around one Behavior per BSID and a single VPP `sr_policy` entry per BSID; the advertised semantics cannot be represented in that model, and silently coercing them (skip Type-A / pick a "winner" behavior) would diverge installed from advertised. The error wording frames it as an agent-side constraint so the controller-side fix (split into per-Behavior BSIDs) is clear.

2. **`BGPPeer.TTLSecurity` propagation** (`calico-vpp-agent/watchers/peers_watcher.go`). Not an RFC violation -- the Calico CR field exists for GTSM (RFC 5082), but the agent never bridges it into `gobgp.TtlSecurity`. VPP's host-eth0 ↔ tap L3 hop consumes the eBGP default TTL=1 (RFC 4271), so external eBGP peers loop on hold-timer-expired until this is bridged. 5-line fix; could be carved into its own PR but kept here because (1) is not testable without it.

### `calico-vpp-agent/routing/bgp_watcher.go`

Two helpers added:

- `walkSRPolicyInnerTLVs(path, fn)`: walks `Path.Pattrs → TunnelEncapAttribute.Tlvs → Tlv.Tlvs`. `path.Pattrs` is heterogeneous, so `UnmarshalTo(TunnelEncapAttribute)` failures are expected (not decode bugs) and are silent-skipped.
- `parseSegmentList(stlv, srnrli) → (types.Srv6SidList, *SegmentTypeB, error)`: per-list parse + validation, returns the trailing `SegmentTypeB` so the caller can read the `EndpointBehaviorStructure`.

`parseSegmentList` rejects (instead of panicking on): empty list, list whose `SegmentTypeB` count > VPP's 16-SID limit, trailing `SegmentTypeB` with nil `EndpointBehaviorStructure`. A list containing `SegmentTypeA` is rejected as a whole list with a distinct error message so operators can distinguish "advertiser sent SR-MPLS we cannot install" from "malformed protobuf at index N".

`getSRPolicy()` is rewritten into a two-level walk and returns `types.SrPolicy.SidLists` with the advertised list count intact.

**Choosing `srv6tunnel.Behavior`.** VPP keeps one Behavior per BSID; the agent uses `srv6tunnel.Behavior` in `connectivity/srv6.go: getPolicyNode` to match a policy against the destination pod prefix family; VPP installs all Segment Lists under one `sr_policy` and ECMP/weight load-balances. Accepting a multi-list policy whose trailing behaviors disagree (e.g. `END_DT4` and `END_DT6`) would install both lists under one BSID, and any packet hashed onto the wrong-behavior list is dropped or mis-decapped at the egress. The PR rejects at receive time and dispatches `SRv6PolicyDeleted` on the existing event bus so any previously-installed policy under the same endpoint follows the same teardown path as a normal withdraw (downstream `SRv6Provider.DelConnectivity` is a pre-existing no-op tracked separately; the event-bus signal here is correct for when that lands).

### `calico-vpp-agent/watchers/peers_watcher.go`

```go
if peerSpec.TTLSecurity != nil && *peerSpec.TTLSecurity > 0 {
    peer.TtlSecurity = &bgpapi.TtlSecurity{
        Enabled: true,
        TtlMin:  uint32(255 - *peerSpec.TTLSecurity),
    }
}
```

Propagates `BGPPeerSpec.TTLSecurity` (GTSM, RFC 5082) into gobgp so external eBGP peers can come up across VPP's extra L3 hop.

### `vpplink/srv6.go`

`AddSRv6Policy` extended for multiple Segment Lists: empty `SidLists` / zero `Bsid` rejected up front, first list via `SrPolicyAdd`, the rest via `SrPolicyMod{Operation: SR_POLICY_OP_API_ADD}`, with a `SrPolicyDel` rollback on intermediate failure so partial state is not left behind. Rollback error message reports 1-based `listIdx/total`.

`sr_policy_add`'s top-level `weight` and `vl_api_srv6_sid_list_t.weight` are wire-format duplicates of the same per-list weight (top-level is authoritative server-side); both are set so the on-wire encoding stays consistent.

## Design decisions

- **Malformed or unsupported-for-this-dataplane input** (no SegmentType decode succeeds / required EBS missing / over the 16-SID VPP compile-time limit / empty list) → reject whole SR Policy.
- **Known-but-uninstallable segment type** (`SegmentTypeA`) → reject list with a distinct error. Priority is "never install a SID chain different from advertised"; skipping Type-A and keeping Type-B would change forwarding behavior.
- **Cross-list trailing-behavior disagreement** → reject whole Policy (implementation-constraint, not RFC-forbidden). Plus dispatch `SRv6PolicyDeleted` for the endpoint so any previous install is signalled for teardown.

### Build / static check

- `make build` (`vpplink`, `calico-vpp-agent`) succeeds
- `go vet ./calico-vpp-agent/... ./vpplink/...` clean
- `gofmt -l` clean
- Unit tests added in `calico-vpp-agent/routing/bgp_watcher_test.go`, cross-compiles with `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go test -c`:
  - `parseSegmentList`: empty list, >16 SIDs, garbage `Any`, Type-A mixed, Type-A only, trailing EBS nil, weight preserve / weight defaults
  - `walkSRPolicyInnerTLVs`: non-TunnelEncap `pattr` silent-skipped, fn-error propagation

### Live cluster

IPv6 single-stack Kubernetes v1.36.1, AlmaLinux 10, kernel 6.12, cri-o 1.32.1, 1 master + 2 workers. External gobgpd VM (`bgp-ctrl`) connected as eBGP peer, injecting SR Policy NLRI (AFI 2 SAFI 73) on the wire.

**Before** (panic + TTL issue both present):

```text
panic: runtime error: index out of range [-1]
github.com/.../routing.(*Server).getSRPolicy
    .../bgp_watcher.go:203
github.com/.../routing.(*Server).injectSRv6Policy
    .../bgp_watcher.go:210

$ kubectl -n calico-vpp-dataplane get pods
calico-vpp-node-v2zb4   1/2   CrashLoopBackOff   872
calico-vpp-node-smtlt   1/2   CrashLoopBackOff   868

$ kubectl ... exec -- gobgp neighbor 192.168.1.13
192.168.1.13 64513 Active |0  0       # hold-timer expired loop
```

**After** (this PR):

```text
$ for n in master worker-1 worker-2; do
    kubectl ... logs $POD -c agent | grep -cE "panic|index out of range"
  done
0 0 0

$ kubectl ... gobgp neighbor
192.168.1.11 64512 Establ |3  3
192.168.1.12 64512 Establ |3  3
192.168.1.13 64513 Establ |1  1
```

**Parser side, multi-list received without panic.** An SR Policy NLRI carrying two `TunnelEncapSubTLVSRSegmentList` entries was injected from bgp-ctrl. All three nodes parsed it and stored `len(SidLists) == 2` in their per-endpoint policy registry without panicking. Agent log:

```text
SRv6Provider new policy cafe::beef with behavior 18 on node fd00:1::11 and priority 0
```

**VPP install + dataplane**, verified on a separate single-Segment-List policy (`cafe::18b`) because the multi-list `cafe::beef` did not match an outbound pod prefix in this run. Triggered by a pod prefix outside `localSidIPPool` (`default-ipv6-ippool` / `fd20::/64`):

```text
$ vppctl show sr policies
[0].- BSID: cafe::18b   Behavior: Encapsulation   EncapSrcIP: fd00:1::12
      Segment Lists:
        [0].- < fcff::11aa:f136:797:4082:31aa > weight: 12

$ vppctl show sr steering-policies
L3 fd20::f136:797:4082:3180/128  cafe::18b
L3 fd20::f136:797:4082:3180/122  cafe::18b

$ kubectl exec te-src -- ping6 -c 5 fd20::f136:797:4082:3182
5 packets transmitted, 5 packets received, 0% packet loss
```

A multi-list `vppctl show sr policies` capture (where `[1].-` is also printed) is on the follow-up list under the interop tests below.

## Compatibility

- Existing BGP UPDATEs with a single `TunnelEncapSubTLVSRSegmentList`: `getSRPolicy` returns `SidLists` of length 1, downstream behavior unchanged (`SrPolicyAdd` only, `SrPolicyMod` never reached).
- VPP already supports `SrPolicyMod { Operation: SR_POLICY_OP_API_ADD }`; no VPP-side patch needed.
- Behavior changes from the previous lenient parser:
  - Trailing EBS missing: old → downstream nil deref panic; new → per-list reject.
  - `SegmentTypeA` in a list: old → silently skipped, only Type-B installed (installed chain ≠ advertised); new → list reject with distinct wording.
  - Cross-list trailing-behavior disagreement: old → walk-order last value silently used (ECMP can land packets on a wrong-behavior list); new → receive-time reject + `SRv6PolicyDeleted` dispatched.
  - Peers without `BGPPeer.TTLSecurity` configured are unaffected (gobgp default `TtlSecurity` disabled).

## Follow-up work

1. Implement `SRv6Provider.DelConnectivity` so normal withdraws and the reject-as-withdraw path can actually remove VPP state.
2. Add more SR Policy interop tests with external GoBGP / FRR controllers (including a multi-list install capture with `[1].-` present in `vppctl show sr policies`).

### DIFF
--- a/calico-vpp-agent/routing/bgp_watcher.go
+++ b/calico-vpp-agent/routing/bgp_watcher.go
@@ -22,6 +22,7 @@ import (
 
 	bgpapi "github.com/osrg/gobgp/v3/api"
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/types/known/anypb"
 	"gopkg.in/tomb.v2"
 
 	calicov3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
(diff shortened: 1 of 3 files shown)


ryskn:
Closing in favor of two focused PRs that split this work along its natural scope boundaries:

- #1022 — BGPPeer.TTLSecurity → gobgp (GTSM bridge, 5 lines)
- #1023 — SR Policy Segment List boundaries + getSRPolicy hardening

The bundled PR was useful for end-to-end validation (GTSM is needed to bring up the external eBGP peer that injects the SR Policy NLRI) but the two changes are independent in code and in review scope. Splitting also lets the small TTLSecurity bridge land independently of the larger SR Policy review.

PR #1023, opened 2026-05-19
fix(bgp,vpplink): preserve SR Policy Segment List boundaries and harden parsing

ryskn:
## Summary

Rework the SR Policy parser and the VPP install path so multi-Segment-List `TunnelEncapSubTLVSRSegmentList` advertisements are preserved end-to-end instead of being flat-concatenated into one strict SID chain, and close several panic exits from `getSRPolicy()`. Prior fixes (#1012, #1018) plugged some panics piecemeal but left the per-Segment-List parse function missing.

(External BGP controller end-to-end validation also needs GTSM on the eBGP peer, filed separately as #1022.)

## Scope

Three categories, not "RFC-violation fixes" wholesale:

- **RFC 9256 alignment.** Stop flat-concatenating multiple `TunnelEncapSubTLVSRSegmentList` entries into one strict SID chain. Independent Segment Lists are the RFC's ECMP / weighted-candidate-path mechanism; flatten silently changes the advertised forwarding semantics.
- **Defensive validation.** Close per-list panic paths (`EndpointBehaviorStructure` nil deref, over-limit count, empty list, garbage `Any`) so malformed advertisements no longer crash the agent.
- **Implementation-constraint rejects.** `SegmentTypeA` (SR-MPLS) presence and cross-list trailing-behavior disagreement are *not* RFC-forbidden, but Calico-VPP's install layer is built around one Behavior per BSID and one VPP `sr_policy` entry per BSID; silently coercing them (skip Type-A / pick a "winner" behavior) would diverge installed forwarding from advertised. Error wording frames these as agent-side constraints so the controller-side fix is clear.

## Changes

`calico-vpp-agent/routing/bgp_watcher.go`:

- `walkSRPolicyInnerTLVs(path, fn)` and `parseSegmentList(stlv, srnrli)` extracted; `getSRPolicy()` rewritten as a two-level walk that returns `types.SrPolicy.SidLists` with the advertised list count intact.
- `parseSegmentList` rejects: empty list, count > VPP's 16-SID limit, trailing `SegmentTypeB` with nil EBS, list containing `SegmentTypeA` (distinct error wording from malformed protobuf).
- Cross-list trailing-behavior disagreement: reject at receive time and dispatch `SRv6PolicyDeleted` on the existing event bus so any previously-installed policy under the same endpoint follows the same teardown path as a normal withdraw. Downstream `SRv6Provider.DelConnectivity` is a pre-existing no-op tracked separately; the event-bus signal is correct for when that lands.

`vpplink/srv6.go`:

- `AddSRv6Policy` extended for multi-list: first via `SrPolicyAdd`, rest via `SrPolicyMod{Operation: SR_POLICY_OP_API_ADD}`, with `SrPolicyDel` rollback on intermediate failure (no partial state in VPP). Empty `SidLists` / zero `Bsid` rejected up front. Per-list `weight` is wire-encoded twice (top-level + nested); both set to the same value.

## Verification

`make build` / `go vet ./calico-vpp-agent/... ./vpplink/...` / `gofmt -l`: clean.

Live cluster (IPv6 single-stack Kubernetes v1.36.1, 1 master + 2 workers, external gobgpd as eBGP peer injecting SR Policy NLRI on the wire, GTSM enabled via #1022):

- Before: `getSRPolicy` panicked on multi-list NLRI and crashlooped the agent (`panic: runtime error: index out of range [-1]` at `bgp_watcher.go:203`).
- After: all three nodes parse the NLRI without panic and store the policy in their per-endpoint registry. A `cafe::18b` policy installed in VPP and Pod-to-Pod ping6 over SRv6 encap (5/5 packets) succeeds end-to-end via a `default-ipv6-ippool` (outside `localSidIPPool`) destination prefix.

A multi-list `vppctl show sr policies` capture (with `[1].-` also printed) is on the follow-up list.

## Compatibility

- Single-Segment-List BGP UPDATEs: `getSRPolicy` returns `SidLists` of length 1, downstream behavior unchanged (`SrPolicyAdd` only; `SrPolicyMod` never reached).
- VPP already supports `SrPolicyMod{Operation: SR_POLICY_OP_API_ADD}`; no VPP-side patch required.
- Behavior changes from the previous lenient parser:
  - Trailing EBS missing: was a downstream nil-deref panic → per-list reject.
  - `SegmentTypeA` in a list: was silently skipped (installed chain ≠ advertised) → whole-list reject with distinct wording.
  - Cross-list trailing-behavior disagreement: was silently resolved by walk-order last value (ECMP could land packets on a wrong-behavior list at egress) → receive-time reject + `SRv6PolicyDeleted` dispatched.

## Follow-up work

1. Implement `SRv6Provider.DelConnectivity` so normal withdraws and the reject-as-withdraw path can actually remove VPP state.
2. Add more SR Policy interop tests with external GoBGP / FRR controllers (including a multi-list install capture with `[1].-` present in `vppctl show sr policies`).

### DIFF
--- a/calico-vpp-agent/routing/bgp_watcher.go
+++ b/calico-vpp-agent/routing/bgp_watcher.go
@@ -22,6 +22,7 @@ import (
 
 	bgpapi "github.com/osrg/gobgp/v3/api"
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/types/known/anypb"
 	"gopkg.in/tomb.v2"
 
 	calicov3 "github.com/projectcalico/api/pkg/apis/projectcalico/v3"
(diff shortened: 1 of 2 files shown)


sknat (approved):
lgtm !

ryskn:
> lgtm !

Thanks for the review!

aritrbas (approved):

PR #1026, opened 2026-05-26
agent: set cnat snat addresses before allowing other servers to run.

hedibouattour:
We saw a bug where we are trying to exclude a prefix before setting cnatsnat entries, which results in NULL entry dereference.

We should have called cnatSetSnat before, which is something we do in handleHostMetadataV4V6Update: where wee send s.GotOurNodeBGPchan <- localNodeSpec before calling s.vpp.CnatSetSnatAddresses(ip4, ip6), so we allowGo(serviceServer.ServeService) & co to run, before setting the cnat snat entry.

The fix is to call cnatSetSnatAddresses before seding the s.GotOurNodeBGPchan signal.


sknat (approved):
lgtm

PR #1029, opened 2026-05-28
watchers: fix route matching to include LinkIndex and Priority

aritrbas:
With multi-uplink configurations on the same physical network, the route watcher creates multiple routes with **identical destination prefix** but **different** `LinkIndex` and `Priority`.
`DelRoute()` and `RTM_DELROUTE` handler matched routes only by `Dst`, ignoring `LinkIndex` and `Priority`.
- `DelRoute`: Matched only by `Dst.String()`, removing the first matching entry from `r.routes` regardless of which uplink's route was actually deleted from netlink.
- `RTM_DELROUTE` handler: Matched only by `Dst.String()`, re-adding the first matching route from `r.routes` even if a different uplink's route was deleted.
 
Updated both locations to match by **`Dst`, `LinkIndex`, and `Priority`**.

### DIFF
--- a/calico-vpp-agent/watchers/uplink_route_watcher.go
+++ b/calico-vpp-agent/watchers/uplink_route_watcher.go
@@ -81,7 +81,9 @@ func (r *RouteWatcher) DelRoute(route *netlink.Route) (err error) {
 		return err
 	}
 	for i, watched := range r.routes {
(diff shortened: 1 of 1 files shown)


sknat (approved):
lgtm

PR #1032, opened 2026-06-01
cni: switch buffer checks to stats segment

aritrbas:
CalicoVPP no longer needs the private `0005-interface-add-buffer-stats-api.patch`. Buffer availability is now read from the VPP stats segment through `govpp` stats client, which was already being utilized by the Prometheus integration.

The agent now creates one shared `statsclient.NewStatsClient()` and passes it to both the Prometheus server and the CNI server. The socket path is configurable through a new `CALICOVPP_STATS_SOCKET` env var and defaults to the `govpp` standard.

This keeps the buffer availability logic aligned with the existing Prometheus stats integration and removes the dependency on the private interface API patch.


sknat (approved):
lgtm

PR #1035, opened 2026-06-02
agent: set cnat snat address before handling ipam pool message

hedibouattour:
**Problem:** Race condition in felix server :
During Felix's initial state sync, IPAMPoolUpdate messages can arrive before the local node's HostMetadataV4V6Update. Both are dispatched in the same single goroutine (the Felix message loop), so this is a deterministic ordering issue, not a data race. The result: CnatAddDelSnatPrefix (called via addDelSnatPrefixForIPPool) fires before CnatSetSnatAddresses has ever been called, causing a NULL dereference in VPP's CNAT SNAT code.

**Fix:** 
 Added a buffer-and-replay mechanism in calico-vpp-agent/felix/felix_server.go:

  1. A pendingSnatPrefix struct holds a pool + isAdd flag.
  2. A snatAddressSet bool field on Server tracks whether CnatSetSnatAddresses has been called at least once.
  3. A pendingSnatPrefixes []pendingSnatPrefix slice queues operations that arrive before the address is set.
  4. A new addDelSnatPrefixForIPPoolOrDefer wrapper replaces the three direct addDelSnatPrefixForIPPool calls in handleIpamPoolUpdate and handleIpamPoolRemove. If the SNAT address isn't set yet, it queues the operation
  instead of calling VPP.
  5. In handleHostMetadataV4V6Update, immediately after CnatSetSnatAddresses succeeds, snatAddressSet is set to true and all queued operations are replayed, then the queue is cleared.

### DIFF
--- a/calico-vpp-agent/felix/felix_server.go
+++ b/calico-vpp-agent/felix/felix_server.go
@@ -110,6 +110,12 @@ type Server struct {
 	ippoolmap  map[string]*proto.IPAMPool
 	ippoolLock sync.RWMutex
 
+	// snatAddressSet is true once CnatSetSnatAddresses has been called for the
+	// local node. SNAT prefix operations that arrive before it are queued here
+	// and replayed immediately after the address is set.
+	snatAddressSet      bool
+	pendingSnatPrefixes []pendingSnatPrefix
+
 	nodeStatesByName  map[string]*common.LocalNodeSpec
 	nodeByWGPublicKey map[string]string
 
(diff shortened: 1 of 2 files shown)


calico-vpp-agent/felix/felix_server.go
```diff
@@ -1209,6 +1215,15 @@ func (s *Server) handleHostMetadataV4V6Update(msg *proto.HostMetadataV4V6Update,
 		if err != nil {
 			s.log.Errorf("Failed to configure SNAT addresses %v", err)
 		}
+		if !s.snatAddressSet {
```
aritrbas: Shouldn't we check for `err == nil` also? --> `if !s.snatAddressSet && err == nil {`
Otherwise, when `CnatSetSnatAddresses` fails, we will end up setting `snatAddressSet = true` and replay all the pending prefixes event though SNAT address wasn't successfully configured in VPP.

sknat: That's a good point, I agree

hedibouattour: right !

aritrbas (approved):

PR #1037, opened 2026-06-04
vpp-manager: bound vpp tarball retention and update clean-vpp

aritrbas:
This patch moves cleanup ownership into `vpp-dataplane` from VPP `kube-test` runner scripts.
- `clean-vpp` now removes local VPP build artifacts in `vpp-manager`.
- also added bounded tarball retention for `vpp-*.tar` to avoid disk pressure. Default retention keeps the latest 12 tarballs.

This keeps the cleanup policy close to the build artifact lifecycle and preserves signal if the build or consumption path regresses.


sknat (approved):
lgtm thanks !

PR #1041, opened 2026-06-05
cni: use govpp buffer stats API for available buffer accounting

aritrbas:
Drop the in-tree VPP private patch `0005-interface-add-buffer-stats-api.patch` (`vpp.GetBufferStats()` via binary API) and read the available buffer count from the VPP stats segment instead. This removes a divergence point from upstream VPP and lets the agent share a single stats client between the CNI buffer admission path and the Prometheus scraper.

* `cmd/calico_vpp_dataplane.go` opens the stats segment exactly once at startup via `core.ConnectStats(statsClient)` and defers a single `Disconnect()` on shutdown. The resulting `*core.StatsConnection` is handed to the CNI server. CNI's `fetchAvailableBuffers()` calls `s.statsConn.GetBufferStats()` directly on every CNI add. Prometheus continues using the same shared `adapter.StatsAPI` directly for `DumpStats()` - the underlying socket is already connected.
* The stats socket path is exposed as `CALICOVPP_VPP_STATS_SOCKET` (defaults to `""` = govpp's standard path).

This requires [FDio/govpp#367](https://github.com/FDio/govpp/pull/367) so that `core.GetBufferStats` accepts `adapter.GaugeStat` as well as `adapter.ScalarStat`. VPP 26.02+ exports `/buffer-pools/<pool>/{available,cached,used}` as `GaugeIndex`, not `ScalarIndex`. Without the govpp fix, every type assertion fails silently and `availableBuffers` is cached as `0` and we observe `Out of buffers: available buffers = 0` errors on every CNI add.

This PR bumps `go.fd.io/govpp` to the pseudo-version `v0.14.0-alpha.0.20260608092747-3c1110da3419` 
([FDio/govpp@3c1110d](https://github.com/FDio/govpp/commit/3c1110da34196e76066e2b29dc41daea175320fe)) after [FDio/govpp#367](https://github.com/FDio/govpp/pull/367) was merged.


hedibouattour (changes requested):
This seems to still report 0 available buffers. I even tested the change before the govpp bump and it seems to be the same case. It always reports 0 no matter what, even though cli shows the good buffers values. I think it's an issue in the govpp api.

hedibouattour:
I went a bit deeper into this issue, it's on govpp side actually.
The type of the buffer stats we are targeting is Gauge, this is shown in vppctl
```
# sh stat seg
...
...
/buffer-pools/default-numa-0/available                                     Gauge           11412    
/buffer-pools/default-numa-0/cached                                        Gauge              86    
/buffer-pools/default-numa-0/used                                          Gauge             458    

```
and govpp 0.13.0 (or 0.11.0 same issue) is trying to parse `ScalarStat`, but not only that, it's also unaware of that type in its mapping, check that in `stat_segment_api.go:54-62`. 
However, we seem to have it added on master (targeting 0.14.0), with this patch:
https://github.com/FDio/govpp/commit/e8ce5754eb389058f7fc25a3f211cf6bf72e3c88
But this alone doesn't fix it, we have to actually consume it.

So, for now :
* either we add the change for `GetBufferStats` to consume the right type, and for that we need to raise the patch against govpp [0]
* Or we do something like `statsConn.GetStatClient().PrepareDir("/buffer-pools/")` and parse things on our own in calicovpp. [1]
And for those two solutions we need to actually target master and not 0.13.0.
(`go.fd.io/govpp v0.14.0-alpha.0.20260528142759-49463705753b`)

Or simpler, we just revert this, going back to the private patch, and do it properly when 0.14.0 is out, including [0].

hedibouattour:
Moving forward with this patch, we can now use the merged change [0] on govpp side by running
`go get go.fd.io/govpp@3c1110da34196e76066e2b29dc41daea175320fe`
[0] https://github.com/FDio/govpp/pull/367
Also, I think prometheus server needs a 
```
       statsConn, err := core.ConnectStats(p.statsclient)
       if err != nil {
               p.log.Errorf("failed to connect to VPP stats", err)
               return nil
       }
       defer statsConn.Disconnect()
```
block to work, in `exportMetrics()`.

aritrbas:
> Moving forward with this patch, we can now use the merged change [0] on govpp side by running `go get go.fd.io/govpp@3c1110da34196e76066e2b29dc41daea175320fe` [0] [FDio/govpp#367]

Thanks @hedibouattour for [FDio/govpp#367] I have updated the patch so that we can now use the merged `govpp` change.

> Also, I think prometheus server needs a
> 
> ```
>        statsConn, err := core.ConnectStats(p.statsclient)
>        if err != nil {
>                p.log.Errorf("failed to connect to VPP stats", err)
>                return nil
>        }
>        defer statsConn.Disconnect()
> ```
> 
> block to work, in `exportMetrics()`.

I tried this, but **it leaks fds**. `statsclient.StatsClient.Connect()` always spawns a fresh `fsnotify.NewWatcher()` (1 inotify fd) and overwrites `sc.done`, orphaning the previous goroutine and its watcher. At the default 5s Prometheus interval, the agent hit `error starting socket monitor: couldn't initialize inotify: too many open files` within seconds.
Instead, `cmd/calico_vpp_dataplane.go` now opens the stats segment **exactly once at startup** via `core.ConnectStats(statsClient)` and defers a single `Disconnect()` on shutdown. The resulting `*core.StatsConnection` is **shared between the CNI server and the Prometheus scraper**.

hedibouattour:
right, lgtm !

hedibouattour (approved):

PR #1042, opened 2026-06-05
Revert "cni: switch buffer checks to stats segment"

aritrbas:
This reverts commit 9f2d2ac1a6371474b2883c54d524c90e4511ce89.

Detailed analysis of the issue from @hedibouattour here: https://github.com/projectcalico/vpp-dataplane/pull/1041#issuecomment-4632613957

I think we should probably revert the culprit commit for now to ensure `master` is not broken.


hedibouattour (approved):
lgtm, thanks !

PR #994, opened 2026-04-30
refactor: unify images into calicovpp/vpp and flatten repo layout

aritrbas:
### Summary:
This PR does two related things on the path to CalicoVPP `release/v3.33.0`:
1. **Unifies dataplane images and flattens the repo layout** - consolidates `calicovpp/agent`, `calicovpp/vpp`, `calicovpp/multinet-monitor` and `calicovpp/init-eks` into a single `calicovpp/vpp` image and moves libraries under `pkg/` with entrypoints under `cmd/`.
2. **Introduces a `kustomize`-based VPP `kube-test` integration** - CalicoVPP now owns the VPP `kube-test` manifest and exposes a small `Makefile` API so VPP `kube-test` (and other consumers) no longer need to know CalicoVPP internal layout details (image names, in-repo VPP build paths, etc.).

**Runtime behavior is unchanged** - the agent and VPP still run as separate containers; the `agent` container reuses `calicovpp/vpp` with an explicit `command: ["/bin/calico-vpp-agent"]` override.

### Changes:
- Unified image: `calicovpp/vpp` (contains all dataplane binaries)
- Single env var: `CALICO_VPP_IMAGE` (replaces `CALICO_AGENT_IMAGE`, `CALICO_VPP_IMAGE`, `MULTINET_MONITOR_IMAGE`)
- `calicovpp/vclsidecar` and `calicovpp/ctl` remain **separate images**
- Unified Dockerfiles (`images/ubuntu/Dockerfile`, `Dockerfile.debug`) ship all binaries: `calico-vpp-agent`, `felix-api-proxy`, `gobgp`, `multinet-monitor`, `vpp-manager` and `init-eks` scripts under `images/ubuntu/bin/`
- Top-level `Makefile` delegates image build through `pkg/vpp-manager`
- `pkg/vpp-manager/Makefile` builds all Go binaries into the unified image
- Per-component `image`/`push` targets removed from `agent` and `multinet-monitor` Makefiles
- Manifests updated to `calicovpp/vpp` with explicit command overrides for `agent` and `multinet-monitor` containers
- Path references in `test/` and `docs/` updated for `pkg/` move
- Image contexts moved to top-level `images/`
- Helper scripts moved to top-level `scripts/`
- VPP clone/build directory moved to top-level `vpp_build/`

### Directory Restructure:
├── cmd
│   ├── api-proxy
│   ├── calico-vpp-agent
│   ├── calicovppctl
│   ├── multinet-monitor
│   ├── replay-trace
[...]

PR #1049, opened 2026-06-19
makefile hardening/cleanup

onong:
Harden/cleanup/fix the makefiles of the many minor issues, oversights and whatnot.


sknat (commented):
looks good, let's maybe wait on #994 to be merged before

sknat (approved):

PR #1051, opened 2026-06-23
prometheus: fix race between metric export and test fetch

aritrbas:
This patch fixes the flaky Prometheus integration tests that started failing after f4674d55e - stats-segment refactor that removed implicit delays (`waitForStatsSocket` + `statsclient.Connect`) which masked a race condition.
- Added `fetchMetricsUntilPresent()` helper that polls `/metrics` until the expected metric name appears, replacing the flaky `fetchMetricsWithRetry()` mehcanism which returned on the first successful HTTP response regardless of content.
- Applied the robust fetch to all tests that assert metric presence (interface stats, TCP stats, pod addition events).
- Also removed the brittle `time.Sleep(2s)` from pod test.


hedibouattour (approved):
thanks !

sknat (approved):
lgtm

PR #1055, opened 2026-07-06
multinet: stop NAD watcher during net watcher cleanup

onong:
cleanExistingWatchers stopped the Network watcher twice and never stopped the NetworkAttachmentDefinition watcher. Stop both watchers so restarts and resyncs do not leave the old NAD watch running.


hedibouattour (approved):
good catch, thanks !

